PR #67, opened 2026-07-17
db/duckdb: manifest fs.read/fs.write -> allowed_directories (mode B)

farkasmark:
Implements DuckDB bounded local file access (design §3.2 mode B): when an app declares fs.read/fs.write, DuckDB read_csv/read_parquet/COPY may touch exactly those directories; otherwise full lockdown (mode A). serve.c resolves the sealed manifest fs paths to absolute dirs into the policy arena; duck_open emits SET allowed_directories + enable_external_access before lock_configuration.

Ordering fix (the default -d connection opened before the manifest loaded, so it couldn't get mode B): added HlAppContextOpts.defer_db_open + hl_app_context_open_db(); serve defers the default open + migrations to after wire_caps installs the policy and before the sandbox seals. The runtime now receives the registry whenever it exists (not only once db_open), so require("hull.db") works under the deferred open.

Verified locally: mode B read within declared dir / blocked outside; mode A blocked; SQLite unaffected; full default suite 59/59; test_db_duckdb 5/5. A new CI step asserts mode B on Linux under the real sandbox.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -802,6 +802,38 @@ jobs:
           cat /tmp/ddbsmoke/out.txt
           grep -q "DUCKDB SMOKE OK" /tmp/ddbsmoke/out.txt
 
+      - name: DuckDB mode B (fs.read -> allowed_directories, named connection)
+        timeout-minutes: 2
+        run: |
+          set -eu
+          # Mode B applies to a named (db.connect) / dynamic (db.open) connection,
+          # which opens after the manifest is known. Under the full pledge/unveil
+          # sandbox, DuckDB read_csv must be able to read a file WITHIN the
+          # declared fs.read directory (which is unveiled). Reads OUTSIDE it are
+          # blocked by the kernel unveil (EACCES); note that DuckDB aborts on a
+          # denied open rather than erroring, so the app must only read declared
+          # paths -- see docs/duckdb_backend_design.md §3.2. The default -d
+          # connection opens before the app loads and stays in mode A (smoke).
+          mkdir -p /tmp/ddbmodeb/data
+          printf 'id,name\n1,alice\n2,bob\n' > /tmp/ddbmodeb/data/test.csv
+          cat > /tmp/ddbmodeb/app.lua <<'EOF'
+          app.manifest({
+              modules = { "hull/db@1" },
+              fs = { read = { "data/*.csv" } },
+              databases = { named = { analytics = "duckdb://:memory:" } },
+          })
+          app.main(function(ctx)
+              local d = require("hull.db").connect("analytics")
+              local rows = d.query("SELECT id, name FROM read_csv('/tmp/ddbmodeb/data/test.csv')")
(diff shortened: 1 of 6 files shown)


farkasmark:
Descoping DuckDB mode B (manifest fs.read/fs.write -> allowed_directories) for now. The mode-B logic works on macOS and under --no-sandbox, but on Linux with the full pledge/unveil sandbox (the production default), a DuckDB file read (read_csv/read_parquet) aborts the process: DuckDB's C++ runtime probes many /sys and /proc paths and NULL-derefs (SIGABRT) when any is denied, and it cannot be reliably enumerated (strace conflicts with pledge's seccomp via rseq). Unveiling /dev/urandom, /etc/localtime, /sys/devices/system/cpu, /sys/fs/cgroup, /proc/self/cgroup, /proc/meminfo and setting fixed memory_limit/threads did not stop the abort. Shipping a feature that crashes under the default sandbox is worse than not shipping it, so mode B stays on this branch for a future focused effort (likely needs upstream DuckDB cooperation or a curated allowlist). The merged state keeps mode A (full lockdown), the sandbox DSN fix (#66), the mbedTLS isolation + --start-group (#65), and the backend + dialect (#64).

farkasmark:
Reopening — the strace failure under the sandbox was a ptrace-vs-seccomp/rseq conflict, not an inherent dead end. Retrying detection with rseq disabled (GLIBC_TUNABLES=glibc.pthread.rseq=0 strace, which runs under the sandbox and shows the actual denials) and with --no-sandbox strace (full file-access enumeration in one pass), so we can unveil the complete set rather than guessing one path at a time.

farkasmark:
Descoping DuckDB mode B — corrected conclusion. After ~15 CI iterations: DuckDB file I/O (read_csv/read_parquet) does NOT work under Hull's Linux pledge/unveil sandbox, and NOT only for undeclared paths — even a read of the DECLARED, unveiled file aborts DuckDB (SIGABRT, internal 'dereference unique_ptr that is NULL'). Earlier evidence suggesting declared reads worked was a misread: the --no-sandbox strace enumeration and the rseq-disabled strace both avoid the sandbox's real runtime behavior (strace's ptrace conflicts with pledge's seccomp: 'Fatal glibc error: rseq registration failed'), so they never exercised the failing path. Attempts that did not fix it: unveiling DuckDB's full probed path set (/dev/urandom, /etc/localtime, /sys/devices/system/cpu, /sys/fs/cgroup, /proc/self/cgroup, /proc/sys/vm/overcommit_memory, /proc/meminfo) and pinning memory_limit/threads to skip auto-detection. This is a fundamental DuckDB-C++-runtime vs restrictive-unveil-sandbox incompatibility, not a bounded allowlist. Mode B works only on macOS (seatbelt) and with --no-sandbox. Merged main keeps the solid pieces: mode A full lockdown, the sandbox DSN fix (#66), the mbedTLS isolation + --start-group (#65), the backend + dialect (#64). Mode B stays on this branch; revisit only with a fundamentally different approach (e.g. running DuckDB work in a separate less-restricted process, or upstream DuckDB sandbox-hardening).

PR #122, opened 2026-07-28
feat(build): HL_ENABLE_IMAGE flavor knob (drop image codecs + stb)

farkasmark:
Adds a subtractive build knob for the image codec subsystem, mirroring `HL_ENABLE_DB`: on by default (web apps want avatars / thumbnails), and `make HL_ENABLE_IMAGE=0` drops it for a compute / CLI / signing binary that never touches images. Removes `cap/image.c` + `cap/image_stb.c` + the per-runtime `mod_image` bindings + vendored `stb_image` (image's **sole** consumer, so stb goes too; ~146 KB smaller on arm64 Darwin).

Classified per the extension taxonomy as a **flavor knob** (subtractive, on-by-default), not a `--with=` feature — the roadmap's near-term-targets table calls this out.

## Wiring (the HL_ENABLE_DB pattern)
- **Makefile:** `HL_ENABLE_IMAGE` flag + `-DHL_ENABLE_IMAGE`; gate `STB_OBJ`; filter `image.c`/`image_stb.c` from `CAP_SRCS` and `mod_image.c` from the per-runtime sources; `IMAGE=` added to the config-sentinel fingerprint; drop `test_image` at 0.
- **Module gate:** new `HL_MOD_CAP_IMAGE` bit; `hull/image` carries it; the resolver's `build_compile_caps` / `build_cap_mask` / `cap_label` learn it — a non-optional `hull/image@1` hard-fails `requires HL_ENABLE_IMAGE (build-time)`, an optional `hull/image@1?` resolves to nil (identical to `hull/gpu@1?`).
- **Source guards:** `mod_image` registration (both runtimes) and the `mod_gpu` image bridge (the `HlImage` texture input + `gpu.texture_read`/`textureRead`, which call `hl_image_new`) behind `#ifdef HL_ENABLE_IMAGE`, so a `GPU=1 IMAGE=0` build keeps raw-byte textures but drops the image bridge. `mod_buffer` needs no guard (it only reads `HlImage` struct fields, which `image.h` still provides).
- **Display parity:** the `build:image` / `image` cap names in `modules.c` and `tool_orchestration.c`; guard the js `import_image` test at `IMAGE=0`.

## Validation
- default `make test` **60/60** (behavior byte-identical), `HL_ENABLE_IMAGE=0` `make test` **59/59** (`test_image` gated)
- the three module scenarios: no-image app runs; non-optional rejected; optional → nil
- stb fully unlinked (0 `stbi_` symbols), ~146 KB smaller
- CI `flavors` matrix gains an `HL_ENABLE_IMAGE=0` link+smoke entry

## Note for review
Built + tested natively (macOS); not yet run through `make CC=cosmocc`. The featurify axis is cosmo-exempt and no platform-branch code was touched, but a cosmo build check before merge would close that out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,6 +235,11 @@ jobs:
             flags: HL_ENABLE_MYSQL=1
           - name: mysql-only (no sqlite)
             flags: HL_ENABLE_SQLITE=0 HL_ENABLE_MYSQL=1
+          # image-less: drops cap/image + stb (image's sole consumer). Guards
+          # the HL_ENABLE_IMAGE=0 link (mod_image/mod_gpu image paths gated) and
+          # that stb is fully unlinked.
+          - name: image-less (no image codecs)
+            flags: HL_ENABLE_IMAGE=0
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5  # v4
         with:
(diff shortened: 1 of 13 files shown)


farkasmark:
✅ **Cosmo build validated** (`make clean && make CC=cosmocc`, cosmocc/gcc 14.1.0).

- APE binary builds + runs (`hull version`); both arch slices present for the image/gpu objects (`build/.aarch64/cap_image.o`, `.aarch64/lua_rt_mod_gpu.o`).
- All image-touched TUs compile under cosmocc with `-DHL_ENABLE_IMAGE`: `cap_image`, `cap_image_stb`, `mod_image` (both runtimes), `mod_gpu` (both runtimes), `module_registry`, `module_resolver`, `tool_orchestration`.
- Cosmo is monolithic (keeps the full base in), so `HL_ENABLE_IMAGE` is always on there and the `#ifdef` guards include all image code → byte-identical cosmo behavior. The knob is native-only, consistent with the other subtractive flags.

Clears the review note above.

PR #146, opened 2026-07-31
feat(tls): auto-compose TLS onto a tls-less base (a2 part 2)

farkasmark:
Completes the **"Keel move"** (docs/tls_feature.md, a2): a plain `hull build`
now composes the mbedTLS stack back onto a tls-less base **only when the app
needs TLS**, and drops it entirely otherwise. This is the compose half of the
a2 base-drop landed in part 1 (#144 / #145).

## The seam and drop
- **release_io** HTTPS client rerouted through the a1 `hl_tls_*` transport seam
  (`hl_tls_client_ctx_create` / `_from_buf` / `hl_tls_config_wire`) instead of
  Keel's `kl_tls_mbedtls_*` directly, so a tls-less base (which excludes Keel's
  `tls_mbedtls.o`) still links release_io's signature-verify path; the weak seam
  fails closed when no TLS is composed.
- The **keel-merge** in the platform lib excludes `tls_mbedtls.o` under
  `HL_TLS_FEATURE=1`, and `libhull_feature-tls.a` carries its own copy
  (extracted + renamed `keel_tls_mbedtls.o` to dodge a member clash) so the
  composed archive is self-contained.
- The pure `hl_crypto_asym_alg_{from,to}_string` helpers + a weak fail-closed
  `hl_cap_crypto_x509_pubkey_pem` move to `cap/crypto.c` (base-resident) so
  `mod_crypto`'s references resolve on a tls-less base where
  `cap/crypto_asym_mbedtls.c` is composed away.

## Build wiring
- `HL_TLS_FEATURE` joins `BUILD_FINGERPRINT` so flipping it auto-purges the
  stale mbedTLS objects + the platform `.a`. HTTP stays on, so the flag has no
  other fingerprint side effect; without this, `make platform HL_TLS_FEATURE=1`
  in a dirty build dir kept a stale mbedTLS `ssl_tls.o` + the full keel merge.
- `build.lua` composes `libhull_feature-tls.a` when the resolved manifest trips
  `needs_tls` **AND** the base lacks `mbedtls_ssl_handshake` (nm probe), so it
  is a byte-identical no-op on a stock TLS-full base.
- Test link sets that pull `release_io.o` gain the transport seam objects
  (`TEST_COMMON_DEPS`/`LIBS`, `test_lua`, `test_release_io`, `test_verify_self`).

## Verification
New CI-wired `tests/e2e_feature_tls.sh`:
- tls-less base links **0** `mbedtls_ssl_handshake`, crypto core + seam intact
- HTTPS app auto-composes mbedTLS back (7 handshake syms)
- tls-free CLI app drops it entirely (**~397 KB** smaller) and runs

`make test` 60/60. `make platform-tlsless` (the mbedTLS-free guard) green.

## Deferred (part 3)
- `release.yml`: fold the `tls` stem into `platform_domain` (`TRUST_FEATURE_LIBS`)
  + the combined sqlite+tls SLIM release wiring.
- Phase 4: retire `pure-compute` into a `build.lua` feature preset now that TLS
  composes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,10 @@ jobs:
         timeout-minutes: 4
         run: make e2e-feature-image
 
+      - name: TLS feature E2E test (tls-free app drops mbedTLS; needs_tls gate)
+        timeout-minutes: 5
+        run: make e2e-feature-tls
+
       - name: Attachment E2E tests
         timeout-minutes: 2
         run: make e2e-attachment
--- a/Makefile
+++ b/Makefile
@@ -456,6 +456,16 @@ HL_TLS_FEATURE     ?= 0
 # unchanged until release.yml opts in. Cosmo ignores it (SQLite stays in-base).
 HL_APP_BASE_SQLITELESS ?= 0
 
+# HL_APP_BASE_TLSLESS=1 (docs/tls_feature.md, a2): the same idea for TLS. The
+# distributed hull embeds a TLS-LESS platform lib as the app-build base (built in
+# an HL_TLS_FEATURE=1 sub-build) plus libhull_feature-tls.a, so a stock `hull
+# build` produces TLS-DROPPING apps (a plaintext app links zero mbedTLS; an HTTPS
+# / net-DB app auto-composes the TLS feature via the needs_tls gate). The hull
+# binary ITSELF stays TLS-full (its own `hull update` needs HTTPS). Composes with
+# HL_APP_BASE_SQLITELESS (both -> a combined sub-build). Only the EMBED_PLATFORM
+# path honours it; a plain dev `make` is unaffected. Cosmo ignores it. Default 0.
+HL_APP_BASE_TLSLESS    ?= 0
+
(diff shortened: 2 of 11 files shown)


farkasmark:
**Discovered (orthogonal, pre-existing follow-up):** while writing the e2e I found that an \`app.main\` CLI app writing to stdout via \`ctx.stdout:write(...)\` **SIGABRTs under the Linux pledge sandbox** — glibc stdio's first-write \`isatty()\` ioctl (\`TCGETS\`/\`TIOCGWINSZ\`) isn't in the CLI \`stdio\` promise allowlist (\`stdio rpath wpath cpath flock fattr\`, no \`tty\`). It's base-independent and TLS-independent (the CLI promise set in \`sandbox.c\` is untouched by this PR), only surfaces on real-Linux landlock+seccomp (macOS + Docker-arm64 mask it), and is rarely hit because Hull apps normally log to stderr via \`log.*\`. The e2e payoff app uses a bare \`return 0\` to stay on-topic. Follow-up: allow the isatty ioctls in the CLI pledge set (or grant \`tty\` for CLI mode) + a regression e2e.

PR #231, opened 2026-08-05
feat(jobs): hull/jobs@1 phase 5 - ops surface + examples + docs

farkasmark:
Phase 5 (final) of `hull/jobs@1` (roadmap §4, design `docs/jobs_design.md`). Stacked on #230 (phase 4). Completes the v1 surface with the operator entry points, a worked example, and a user guide.

## Ops surface (both runtimes)

- **`jobs.dead(opts)`** — list dead-lettered jobs (`status='dead'`), newest first, with `last_error` + timestamps. `{ queue, limit=100, offset=0 }`.
- **`jobs.retry(id)`** — requeue a dead job: reset to `pending` with a fresh attempt budget, clear the error. No-op unless the row exists and is currently `dead` (can't double-requeue a live job). Returns whether it requeued.
- **`jobs.cleanup(opts)`** — purge terminal (`done` + `dead`) rows past a retention age (`older_than`, default 7d; or an absolute `before` ts; `statuses` overridable). Only touches terminal rows, so it never races a live job. Returns the deleted count.

## `examples/jobs`

A runnable demo in **both runtimes** showing both execution models:
- **In-process poller** (`app.lua`/`app.js`) — an `app.every` timer drives `jobs.work`, with `/jobs` enqueue + `/jobs/stats` + `/jobs/dead` + `/jobs/retry/:id` routes and a nightly `app.daily` cleanup.
- **Dedicated worker** (`worker.lua`/`worker.js`) — `app.main` runs `jobs.run_worker`; run K copies against the same DB.

Plus `examples/jobs/README.md` and a full user guide at **`docs/jobs.md`**.

## Tests

`e2e_jobs.sh` gains a phase-5 check in both runtimes: two jobs dead-letter, `jobs.dead` lists them (with `last_error`), `jobs.retry` requeues one (and no-ops on a bad id), `jobs.cleanup` purges the remaining terminal row while leaving the requeued pending job. **14/14 green** locally. Both example apps verified end-to-end (poller drains; worker drains; `hull jobs worker` launcher).

## Note

The JS example poller uses the **documented fire-and-forget** timer form (`app.every(ms, () => jobs.work())`). While building the example I found that an `async` timer callback that `await`s `jobs.work` re-arms unreliably — an `app.every`/async interaction, **not** a jobs bug (`run_worker` and the sync `work` path are unaffected, and the design docstrings already recommend the fire-and-forget form). Flagging as a separate follow-up.

## Epic complete

Phases 1–5 land the full `hull/jobs@1`: schema + atomic claim (SQLite/PG/MySQL) + handlers/work-loop/retry/dead-letter/reaper + dedicated worker + ops. Ready to collapse the stack (#227→#228→#229→#230→#231) into main after review.


farkasmark:
## Audit + API review (folded into this PR)

Ran `/c-audit` `/lua-audit` `/js-audit` on the jobs surface plus a usability/flexibility/performance review. **C** (`cmd_jobs.c` + the `supports_skip_locked` dialect flag) and **Lua** came back clean — all SQL is parameterized, including the dynamically-built `IN (?,?)` placeholder lists (only the `?` string is concatenated; values bind). One real JS finding and two gaps, all fixed in the latest commit:

- **J1 (JS, Medium — correctness + Lua/JS parity):** `_handlers` was a plain object, so `_handlers[job.type]` inherited from `Object.prototype`. A job whose `type` matches an inherited name (`toString`, `constructor`, `valueOf`, …) resolved to a function/value instead of `undefined`, **bypassing the no-handler → default/dead path** (silently `markDone` or mis-dispatched). Reachable when the app enqueues a user-controlled type — which `examples/jobs` `POST /jobs` does. Not RCE (QuickJS disables `Function`). Fixed with `Object.create(null)`; Lua tables have no prototype chain and were already safe. Verified: a `toString` job type now dead-letters (`done=1 dead=2`).
- **`jobs.cancel(id)`** (both runtimes) — delete a still-`pending` (delayed/scheduled) job; no-ops on a running/missing one. Closes the obvious usability gap: `cleanup` only purged *terminal* rows, so there was no way to un-schedule a queued job.
- **Reaper throttle** (both runtimes) — `jobs.work` reaped on *every* call; a no-op sweep still takes the write lock (SQLite is single-writer), so a fast poller × N workers added needless contention. Now throttled to `reap_interval` (default 30s; `0` = every call). Correctness unchanged — an orphan is reclaimed within `visibility_timeout + reap_interval`.

Also: a `.catch` guard on the example JS poller, and stale phase-reference comments cleaned. `e2e_jobs.sh` phase-5 check extended with `cancel`; **14/14 green**.

Left as-is (flagged): `stats.failed` is a permanently-0 field (`dead` is the failed-terminal state) — kept because the design doc documents the shape, but it's a minor wart we could drop pre-release.

Issue #232, opened 2026-08-05
hull/jobs@1 v1.x follow-ups: cron, intra-process concurrency, heartbeat, rate-limiting, low-latency wakeup


farkasmark:
Follow-ups surfaced by the post-merge API review of `hull/jobs@1` (merged in #231), benchmarked against the DB-backed peer group (pg-boss, Graphile Worker, River) and the broader bar (Celery, BullMQ, RQ, Dramatiq).

The durable core shipped and is competitively complete: atomic claim (SKIP LOCKED / SQLite-serialized), retry+backoff, dead-letter + inspect/requeue, visibility-timeout reaper, delay/priority/named-queue/dedup, cancel, cleanup, two execution models, transactional enqueue (the headline differentiator vs every Redis broker), backend-agnostic (SQLite/PG/MySQL — unique vs PG-only pg-boss/Graphile/River).

These are the gaps, ranked by how prominently the field has them:

### 1. Cron / recurring jobs  ⭐ (top usability gap)
Everyone else has first-class cron (Celery beat, BullMQ repeatable, pg-boss schedule, Graphile crontab). We defer to "compose `app.every`". A `jobs.cron(spec, type, data)` (5-field cron) that enqueues on schedule would close the single most-noticed gap. **Design fork:** durable DB-backed (`_hull_cron` table, survives restart, multi-worker-safe — matches the durable ethos) vs in-process registration.
→ **Targeted in `feat/jobs-v1.1`.**

### 2. Intra-process concurrency  ⭐ (top performance gap)
`run_worker` dispatches handlers **sequentially**, so one worker = one in-flight job even for I/O-bound async handlers; you only parallelize by running K processes. A `concurrency: N` option (N in-flight handlers per worker) would use each process far better. JS is easy (`Promise.all`); **Lua needs a small runtime primitive** — promote the existing detached-spawn `tui.async` to a general `hull.async` (the code comment in `mod_tui.c` already anticipates this).
→ **Targeted in `feat/jobs-v1.1`.**

### 3. Long-job heartbeat / lock extension  (correctness sharp-edge)
A handler that legitimately runs longer than `visibility_timeout` gets reclaimed by the reaper and **double-run**. Today's only lever is a high `visibility_timeout`. A `job:touch()` / lock-extend (or auto-heartbeat while a handler runs) removes the footgun. At minimum needs a loud doc callout now.

### 4. Rate limiting / throttling
[...]

PR #233, opened 2026-08-05
feat(jobs): v1.1 - durable cron + intra-process concurrency

farkasmark:
Closes the top two follow-ups from the post-merge API review (#232): **cron** (top usability gap) and **intra-process concurrency** (top performance gap). Design forks were confirmed: durable DB-backed cron + concurrency in both runtimes.

## Runtime primitive: `hull.async`
Promotes the detached-spawn `tui.async` to a general `hull.async(fn)` (`async.c`); `tui.async` now aliases it. Lua stdlib couldn't drive concurrent async coroutines without this — the async-resume machinery needs C-level `active_co` bookkeeping that plain `coroutine.create` doesn't set. Dedups ~50 lines of tricky coroutine code out of `mod_tui.c`.

## Intra-process concurrency (both runtimes)
`run_worker({ concurrency: N })` runs N independent claim-loops in one process → up to N handlers in flight at once (real parallelism for I/O-bound handlers; each loop claims disjoint jobs via the atomic claim). JS joins with `Promise.all`; Lua spawns N-1 detached `hull.async` loops + 1 inline and waits on a shared `active` counter. Orthogonal to K processes — they multiply.

## Durable cron (both runtimes)
`jobs.cron(name, spec, data?, opts?)` + `jobs.uncron(name)`. Schedules live in a new `_hull_cron` table (survive restarts, coordinate across the fleet) and fire via a **compare-and-set on `next_run_at`** — exactly one worker fires each tick on **every** backend (no SKIP LOCKED needed). Standard 5-field cron (UTC), parsed + next-occurrence computed in **pure Lua/JS** via the civil-calendar algorithm (no `gmtime` dependency; coarse-to-fine day/hour/minute skips keep it fast even for `0 0 29 2 *`). Fired from `work()` (throttled 1s). Missed ticks advance to the next future occurrence (fire-once, no backfill storm).

Also fixed: a NULL `max_attempts` override on the JS cron→enqueue path was binding NULL into the NOT-NULL column (Lua's `or` already handled it).

## Tests
`e2e_jobs.sh` +concurrency (N in-flight, exactly-once) and +cron (`cron_next` math on a fixed epoch, fire, no-double-fire via CAS, uncron) in both runtimes — **18/18 green**. Full unit **62/62**; `test_lua`/`test_js` + TUI e2e (validating the `tui.async` dedup) all green.

Refs #232. Remaining follow-ups there: long-job heartbeat (correctness), rate limiting, LISTEN/NOTIFY low-latency.

### DIFF
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -164,6 +164,41 @@ jobs.cleanup({ older_than = 7 * 86400 })   -- purge terminal (done + dead) rows 
 ```
 `jobs.cancel` only removes a `pending` job (a delayed/scheduled one that hasn't
 started); a `running` job is mid-flight and is left to finish or dead-letter.
+
+## Compute-heavy jobs (WASM / GPU)
+
+A handler is ordinary app code, so it has full capability access - `compute.*`
+(WASM), `gpu.*`, `db.async.*`, `http.fetch`, everything the app's manifest
+grants. Offloading heavy work to a background job is a natural fit. Five things
+to get right (these are general compute-orchestration rules, not jobs-specific,
+but the visibility timeout makes the last one sharper):
+
+- **Use the async variants.** `compute.async.call` / `gpu.async.dispatch` yield
+  to the event loop; `compute.call` / `gpu.dispatch` **block** the single
+  event-loop thread for the whole computation, stalling every other concurrent
+  loop, timer, and (if the process also serves) HTTP request. With
+  `run_worker({ concurrency = N })` this matters more - one sync compute call
+  freezes the other N-1 loops.
+- **Parallelism is bounded by the thread pool, not by `concurrency`.**
+  `compute.async` / `gpu.async` / `db.async` all dispatch to one shared worker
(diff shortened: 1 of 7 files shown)


farkasmark:
Added two more #232 follow-ups to this branch:

- **`jobs.get(id)`** (both runtimes) — one job's full status view (or nil), the "did my job run?" primitive. Closes the clearest usability gap: job-by-id was previously impossible from app code (`_hull_jobs` is namespace-protected). `shape_ops` now also carries status/priority/run_at.
- **`jobs.heartbeat(job)`** (both runtimes) — extend the claim on a job the current handler is processing so a long handler isn't reclaimed and re-run. Guarded by the job's `claim_token` (now attached to each claimed job); returns `false` once the claim is lost — the handler's anti-double-run signal. Removes the long-compute-job footgun.

`e2e_jobs.sh` now **20/20** (added get/heartbeat in both runtimes). This branch (v1.1) now lands cron + concurrency + get + heartbeat.

PR #259, opened 2026-08-06
jobs: strict per-key concurrency (counter-backed hard cap)

farkasmark:
## Strict per-key concurrency (counter-backed hard cap)

Completes the "soft now, strict later" plan from #256. Opt into a **hard cap** with `concurrency_strict = true` (JS `concurrencyStrict: true`) on enqueue:

```lua
jobs.enqueue("charge", data, { concurrency_key = "acct-7", concurrency = 1,
                               concurrency_strict = true })   -- serialize per account
```

### Soft vs strict
- **Soft** (default, #256): reconciles a derived rank after the claim; may briefly overshoot under same-instant races. Right for fairness/politeness caps.
- **Strict** (new): reserves a slot in a fleet-wide counter (`_hull_job_concurrency`) **atomically at claim** (`n < limit -> n+1`), under the same blocking-`FOR UPDATE` / SQLite-write-txn serialization as `jobs.limit`. N racing workers admit **exactly** the limit, never more.

### Lifecycle
- Slot released whenever a job leaves `running` (done / dead / retry / workflow yield).
- The reaper **reconciles** the counter to the true running count every sweep, so a crashed worker can't permanently leak a slot — it self-heals, briefly under-utilizing until the next reap (the safe direction for a hard cap).

### Schema
One nullable column (`concurrency_strict`) + one counter table, both added idempotently for existing DBs (`ensure_column` / `CREATE TABLE IF NOT EXISTS`).

### Tests
- Single-process cap+drain (both runtimes).
- **Fleet gate**: K parallel claimers on one strict key claim *exactly* the limit — the no-overshoot guarantee soft can't make.
- **Validated locally on all three backends**: SQLite (58/58 e2e), MySQL 8, PostgreSQL 16 (Docker). The reserve, release, and the correlated-subquery reaper reconcile all run cleanly on each (the reconcile subquery reads a different table, so it's MySQL-legal — checked deliberately after the #257 `LIKE ESCAPE` lesson).

First item of the "finish hull/jobs polish" pass.

### DIFF
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -202,14 +202,30 @@ jobs.enqueue("report", data, { concurrencyKey: "tenant-42", concurrency: 2 });
   of that key finishes, a held-back one becomes claimable. Priority and FIFO
   order within the key are preserved (the release is ranked by claim age, so it
   is deterministic fleet-wide - no two workers evict the same slot).
-- **Soft** semantics: correct under normal load, but two workers claiming the same
-  key in the same instant can briefly exceed the limit (there is no reservation
-  counter). Use it for *fairness / politeness* caps (don't let one tenant hog the
-  fleet, cap per-host outbound), not as a hard mutual-exclusion lock. A strict,
-  counter-backed variant is a planned opt-in follow-up.
-- Costs one small indexed DB read per claimed job **only for keyed jobs**;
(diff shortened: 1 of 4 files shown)


farkasmark:
Superseded by the rebased-onto-events version (see the new PR from strict-on-events). Same commit, applied on the merged #268 base for a clean CI run.

Issue #246, opened 2026-08-05
jobs v1.6: durable, fleet-wide lifecycle events


farkasmark:
## jobs v1.6: durable, fleet-wide lifecycle events

**Context.** v1.5 (#244) shipped in-process lifecycle hooks: `jobs.on("completed"|"retried"|"dead", fn)`, fired synchronously inside `jobs.work` **in the worker that processed the job**. That covers per-worker observability, metrics, and side-effects. It intentionally does **not** deliver events across processes: a separate dashboard/reactor process sees nothing about jobs run by other workers.

This issue tracks the **durable, cross-process** complement — the "any worker ran a job, every subscriber hears about it" story.

### Motivation
- A dedicated observability/dashboard process that reacts to `dead` jobs fleet-wide (alerting, DLQ triage).
- A reactor process that fans out on `completed` without being the worker that ran the job.
- An audit/event log of terminal transitions that survives restarts.

### Relationship to #235 (LISTEN/NOTIFY)
Overlapping transport. On **Postgres**, `LISTEN`/`NOTIFY` is the natural low-latency delivery: `NOTIFY` on each terminal transition, subscribers `LISTEN`. On **SQLite/MySQL**, a durable events table + polling. So this should likely be **built on #235's primitive** where available, with a portable table-backed fallback — i.e. #235 delivers the wakeup, this defines the event payload/API and the durable log. Decide whether to unify or keep separate.

### Design sketch (options, not decided)
- **Durable event log**: `_hull_job_events(id, job_id, event, info, created_at)` appended on each terminal transition; a cursor API `jobs.events({ since, limit })` and/or a subscription `jobs.subscribe(events, fn)` that polls (or LISTENs on PG). Retention/cleanup like `jobs.cleanup`.
- **Transport**: PG `LISTEN`/`NOTIFY` (via #235) for low latency; table-poll fallback elsewhere.
- **API**: keep the in-process `jobs.on` (per-worker, synchronous) distinct from a fleet-wide `jobs.subscribe`/`jobs.events` (durable, cross-process). Naming TBD.

### Open questions
1. Delivery guarantee — at-least-once + subscriber-side dedup (cursor/offset)? Exactly-once is out of scope (mirrors the queue's own at-least-once).
2. Retention/cleanup policy for the event log.
3. Unify with #235 or ship as a layer on top of it.
4. Which events — the three terminal ones (`completed`/`retried`/`dead`), or also `enqueued`/`claimed`?
[...]

PR #262, opened 2026-08-06
jobs: durable fleet-wide events, Phase 1 (log + transactional emit + tail)

farkasmark:
## Durable fleet-wide events — Phase 1 (log + transactional emit + tail)

First phase of #246 (design: `docs/jobs_events_design.md`, RFC #261). Where `jobs.on` is in-process and lost on a crash, this adds an **opt-in durable event log**: `jobs.init({ events = true })` records every lifecycle transition (`enqueued` / `completed` / `retried` / `dead` / `cancelled`) as a row in `_hull_job_events`.

### The keystone: transactional coupling
The event row is written in the **same transaction** as the state change that produced it. A new `finish()` helper wraps each terminal outcome (`mark_done`/`mark_dead`/`mark_retry`) + the durable emit in one `db.batch`, so an event exists **iff** the transition committed — no lost events, no phantom events. As a bonus it collapses `mark_done`'s several statements into one commit (fewer fsyncs), a **net win even when events are off**. `enqueue` emits `enqueued` inline (joining the caller's `db.batch` when they wrap it); `cancel` emits `cancelled`. `emit_durable` no-ops when events are off, so the default path is unchanged.

### The tail
```lua
jobs.init({ events = true })
local recent = jobs.events({ types = { "dead" }, since = last_id, limit = 100 })
-- each: { id, ts, type, job_id, job_type, queue, data = { error?, attempt?, trace? } }
```
`id` is the monotonic total order (+ the future subscription cursor space); `data` is compact metadata, never the full result. `jobs.cleanup` retains the log by age. **Off by default** — an app that doesn't opt in pays nothing.

### Tests / validation
- e2e asserts the event counts match committed state **exactly** (no phantom events) + the type filter; the events-off no-op is covered by all 55 existing tests (which run events-off through the new `finish()` path).
- **SQLite 57/57 + MySQL 8 + PostgreSQL 16** (Docker) — the transactional `finish()` wrapping and the log/tail SQL run cleanly on each (validated locally, the `LIKE ESCAPE`/correlated-subquery discipline).
- Both runtimes, full parity. luacheck clean.

Ships value alone: a dashboard can poll `jobs.events`. Phase 2 (durable subscriptions with cursors + at-least-once delivery) is the next step.

### DIFF
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -113,6 +113,42 @@ jobs.on("completed", function(job, info) metrics.inc("jobs.done", { type = job.t
 jobs.on("dead",      function(job, info) log.error("job died: " .. info.error) end)
 ```
 
+## Durable events (fleet-wide)
+
+Where `jobs.on` is in-process and lost on a crash, the **durable event log** is
+opt-in (`jobs.init({ events = true })`) and records every lifecycle transition -
+`enqueued`, `completed`, `retried`, `dead`, `cancelled` - as a row in
+`_hull_job_events`, **written in the same transaction as the state change that
+produced it**. So an event exists *iff* the transition committed: no lost events
+(crash after the change, before an external publish) and no phantom events
+(publish, then the change rolls back) - the same transactional coupling that makes
+`enqueue` a plain `INSERT`. It survives restarts and is visible to any process
+against the same DB (a dashboard, an analytics sink, a reacting service).
+
+```lua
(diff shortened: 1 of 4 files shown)


farkasmark:
Superseded by #268 (the three events phases collapsed into one squash for a clean fresh-CI merge; the originals never got CI due to the 2026-08-06 outage).

Issue #235, opened 2026-08-05
epic: jobs low-latency wakeup via Postgres LISTEN/NOTIFY


farkasmark:
Split out from #232 as its own epic — unlike the other `hull/jobs` follow-ups (which were stdlib-only), this is a **C-level, Postgres-only** change.

## Goal
Reduce a worker's **idle-wakeup latency** (time to pick up the first job after an idle period) from up to `poll_ms` (default 1s) to ~milliseconds, using Postgres `LISTEN`/`NOTIFY`. Matches Graphile Worker's sub-ms pickup.

## Scope note (why it's lower priority than it looks)
- **Latency-only, not a correctness/capability gap** — polling is already correct.
- `jobs.run_worker` already **busy-drains**: it loops and re-claims immediately while jobs exist, sleeping `poll_ms` **only when the queue is empty**. So steady-state latency under load is already low; NOTIFY only helps the *idle → first job* transition.
- **Postgres-only.** SQLite and MySQL keep polling (SQLite has no LISTEN/NOTIFY; MySQL has none). So this is a per-backend fast path layered under the existing poll loop.
- Interim lever available today: lower `poll_ms` (trades a bit of DB read load for latency).

## Work breakdown

**1. pg wire client — async `NotificationResponse` ('A').**
`src/hull/cap/pgwire.c` / `pg_conn.c` currently have no handling for the async `NotificationResponse` ('A') message, which the server can send at any time after `LISTEN`. Add parsing + a way to surface pending notifications to the caller. (Also handle 'A' arriving interleaved with normal command responses.)

**2. DB-cap primitive — event-loop-integrated wait.**
The hard part. Add a capability like `conn.listen(channel)` + `conn.wait_notify(timeout_ms)` that **yields to Hull's async event loop** (registers the pg socket for read-readiness and resumes the coroutine on a notification or timeout), rather than blocking a pool thread — mirroring how `http.fetch` / `db.async` integrate with the async backend (`hl_async_*`). This is the bulk of the design.
- Needs the pg connection's fd exposed to the event loop.
- Fail-closed + a clean fallback to plain sleep when the connection isn't a live socket (e.g. the sync path, or a non-PG backend).

**3. jobs wiring (stdlib).**
- On `jobs.enqueue` (and cron fire), `NOTIFY hull_jobs_<queue>` — cheap, and only meaningful on PG.
[...]

PR #263, opened 2026-08-06
jobs: durable events Phase 2 (subscriptions - cursor + lease drain)

farkasmark:
## Durable events — Phase 2 (subscriptions: cursor + lease drain)

Second phase of #246, stacked on Phase 1 (#262). Adds durable, fleet-wide event **subscriptions**: `jobs.subscribe(name, handler, {types, from})` / `jobs.unsubscribe`. Each event is delivered **at-least-once**, in id order, with the subscription tracking its own durable cursor (`_hull_job_subscriptions`) so it resumes exactly where it left off after a restart.

### Fleet-safe, no global lock
A per-subscription **lease** (claim-token + visibility timeout — the same mechanism as the job claim/reaper) means one worker drains a subscription at a time; a crashed drainer's lease expires and another resumes from the unadvanced cursor. Drained by `jobs.work` / `run_worker` (throttled; only workers that registered subscribers do work). Retention is now **subscription-aware** — `jobs.cleanup` never truncates past `min(cursor)`.

### Testability keystone
`jobs._events_drain(name, {now, batch, commit_cursor, release_lease})` — a **synchronous, clock-injectable** seam (no timers/sleeps). Every guarantee is asserted at exact instants:
- deterministic in-order capture;
- crash-before-cursor → re-delivery (`commit_cursor=false`);
- worker-death lease reclaim (`release_lease=false` + advanced `now`);
- retention-never-past-`min(cursor)`;
- plus a real multi-process **fleet gate** (K `hull` processes drain one subscription → every event delivered exactly once).

### Cross-backend
Both runtimes, full parity. **SQLite 60/60 + MySQL 8 + PostgreSQL 16** (Docker). The 3-backend pass caught three portability issues, all fixed:
- `cursor` is a **MySQL reserved word** → `cursor_id`;
- the lease CAS must detect acquisition via `UPDATE … RETURNING` on PG/SQLite because **`db.exec` doesn't surface an affected-row count on Postgres** (MySQL returns it);
- the test's cleanup cutoff must stay within int4 (second-resolution `ts` columns are INTEGER jobs-wide, per convention).

**Discovered follow-up (pre-existing, out of scope):** the same pg `db.exec`-rowcount gap means `jobs.cancel`/`retry`/`reap` report wrong *counts* on Postgres (the statements still execute). Worth a pg-backend fix.

Phase 3 (failing-subscriber skip/backoff + lag metrics) and Phase 4 (LISTEN/NOTIFY, #235) are next.

### DIFF
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -137,17 +137,46 @@ jobs.init({ events: true });
 const recent = jobs.events({ types: ["dead"], since: lastId, limit: 100 });
 ```
 
-- **`id`** is a monotonic total order (and the cursor space Phase 2 subscriptions
-  will use). Tail incrementally by passing the last id you saw as `since`.
+- **`id`** is a monotonic total order (and the subscription cursor space). Tail
+  incrementally by passing the last id you saw as `since`.
 - **`data`** is compact metadata (`error` / `attempt` / `trace`), never the full
   result - join `jobs.result(job_id)` when you need the value.
-- **Retention** is by age via `jobs.cleanup` (same window as terminal rows).
 - **Off by default** - an app that doesn't opt in pays nothing (no log writes).
-- This is **Phase 1** (the log + a pollable tail, enough to power a dashboard).
-  Phase 2 adds **durable subscriptions** (`jobs.subscribe(name, handler)`) with
-  at-least-once delivery + per-consumer cursors, so a reacting service processes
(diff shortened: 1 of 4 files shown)


farkasmark:
Superseded by #268 (the three events phases collapsed into one squash for a clean fresh-CI merge; the originals never got CI due to the 2026-08-06 outage).

Issue #295, opened 2026-08-09
build: hull build --with=<feature> fails to link against an ASan/UBSan (make debug) build tree


farkasmark:
## Summary

`hull build --with=<feature>` (feature composition) fails at the final link when
the local `build/` tree was produced by `make debug` (ASan+UBSan). The composed
`libhull_platform.a` / feature archive contain sanitizer references
(`___ubsan_handle_*`, ASan interceptors), but the app-link that `hull build`
drives through the default compiler-free emitter or a plain system `cc` does NOT
pass `-fsanitize=...`, so those symbols are undefined:

```
Undefined symbols for architecture arm64:
  "___ubsan_handle_sub_overflow", referenced from:
      _b64url_decode_char in libhull_platform.a[6](cap_crypto.o)
  ...
ld: symbol(s) not found for architecture arm64
hull build: linking failed
```

## Scope

This is a **pre-existing, feature-agnostic** property of the compose path, not
specific to any one feature: it reproduces with `--with=duckdb` / `--with=postgres`
/ `--with=mysql` / `--with=gpu` / `--with=valkey` equally, because the platform
lib (not the feature) is the instrumented input. The release / CI feature-compose
e2e jobs are unaffected (they build non-instrumented `EMBED_PLATFORM=1` trees), so
this only bites a developer who has a lingering `make debug` sentinel in `build/`
and then runs `hull build --with=`.

## Workaround

`make clean && make EMBED_PLATFORM=1` (a non-instrumented tree) before composing.

## Possible fixes (build-system, out of scope for any feature PR)

- Have `hull build` detect an instrumented platform lib (e.g. a marker in
  `build/.build-config`, or an `nm` probe for `___ubsan_`/`__asan_`) and either
  refuse with an actionable message or auto-thread the sanitizer runtime into the
  app link.
- Or record the sanitizer state in the composed-feature metadata and surface a
  clear "this build/ tree is instrumented; recompose from a release tree" error
  instead of a raw linker dump.

Surfaced while validating the Valkey/Redis `--with=valkey` feature; filed
separately per review guidance rather than expanding that PR.

PR #296, opened 2026-08-09
feat(kv): Valkey/Redis as the --with=valkey composable KV feature

farkasmark:
Adds Valkey/Redis as a composable `--with=valkey` feature backing the shipped
milestone-1 `hull/kv` and `hull/cache` APIs (`backend = "valkey"`), across both
runtimes. It is the **first non-SQL connection feature**: Redis is a command
protocol, not a query language, so it does not go behind `HlDbBackend` — it
introduces a new narrow `HlKvBackend` seam (`include/hull/cap/kv_backend.h`) with
its own weak hook `hl_kv_feature_backends` (`cap/kv_feature.c`), a sibling of
`hl_db_feature_backends` so the two compose side by side.

Scope is the portable KV subset only. There is **no** generic Redis-command
escape hatch; cluster / Sentinel / pub-sub / streams / scripting / data-structures
are out of scope.

## Layers

- **RESP2/RESP3 codec** (`cap/respwire.c`) — bounds-checked over untrusted input
  (depth/bulk/element caps), binary-safe, glob-escaping writer for SCAN.
- **Connection** (`cap/valkey_conn.c`) — raw blocking socket (getaddrinfo /
  connect / SO_RCVTIMEO+SNDTIMEO), `HELLO 3` with RESP2 fallback, ACL/password
  auth, `rediss://` TLS via the shared `shared/tls_client.c` (embedded CA bundle;
  Keel is inbound-only, not used here). Pluggable `HlAllocator` + `sh_arena` reply
  arena; zero raw malloc/free; DSN password scrubbed post-handshake.
- **Backend vtable** (`cap/valkey.c`) — get/set/del/exists, TTL, atomic incr
  (transactional WATCH/MULTI/EXEC when a TTL lands on a fresh key so an existing
  key keeps its TTL), CAS (SET NX / WATCH-MULTI-EXEC → four-state
  OK/MISMATCH/CONFLICT/ERROR), bounded cursor SCAN + clear (never blocking KEYS,
  iteration-capped).
- **App cap + bindings** — base-resident `cap/kv.c` (select → open → dispatch with
  cap-gated "unsupported"); `runtime/{lua,js}/mod_kv.c` (`hull.kv._native`);
  stdlib `hull.kv._valkey` (Lua+JS) wiring `kv.open`/`cache.open`.
- **Manifest + sandbox** — `kv = { dynamic = { schemes, hosts } }` (`cap/kv_dynamic.c`
  validator, mirrors `databases.dynamic`); the sandbox grants `network_outbound`
  for a declared KV net policy.
- **Packaging** — `hull build --with=valkey` / `hull feature install valkey`;
  `feature_specs.lua` + `FEATURES[]` + `mk/features/valkey.mk` + release matrix.

## Security properties

- **Borrow-copy guard.** `get()` returns a value borrowed into the connection's
  receive buffer; both bindings copy it into runtime-owned memory before
  returning (`lua_pushlstring` / `JS_NewArrayBufferCopy`), so a script value
  survives any later op on the same connection. Regression-tested end to end
  (get → large-value op that reuses the buffer → first value unchanged).
- **Namespace isolation.** Physical keys are `kv:<hex(ns)>:` / `cache:<hex(ns)>:`;
  hex-encoding the namespace makes prefixes collision- and SCAN-glob-free, so a
  `kv` and a `cache` sharing a namespace name (or two namespaces where one is a
  prefix of the other) never alias on the shared server. Byte-for-byte identical
  across Lua and JS.
- **Allowlist, fail-closed.** `kv.open` is denied unless the DSN scheme is in
  `kv.dynamic.schemes` and the host matches `kv.dynamic.hosts` (exact / `*.suffix`
  / CIDR / `$VAR`, the shared matcher); a null/undeclared/empty policy denies
  everything. Enforced before any connect.
- **Sandbox.** A declared KV net policy grants `network_outbound` (else the
  connect is SIGKILLed on Linux / EPERM on macOS); no other authority is opened.
- **Injection-safe wire + bounded parsing** over untrusted server input; DSN
  password scrubbed after the handshake; a plaintext `redis://` with a password
  warns once toward `rediss://`.
- **Base stays clean.** A non-valkey base links **zero** backend symbols
  (`nm base | grep hl_kv_backend_valkey` → empty); the archive omits the
  self-registration TU (`cap/valkey_register.c`) so the generated registry is the
  sole hook provider (no duplicate symbol, no build-time macro).

## Test results (local, this branch)

Unit (utest, ASan+UBSan) — **all green**:

| suite | cases |
|---|---|
| test_respwire | 23 |
| test_valkey_dsn | 18 |
| test_valkey_conn | 7 (socketpair) |
| test_valkey_backend | 3 (socketpair) |
| test_kv_dynamic | 3 |

Full unit suite: **67/67 suites pass**. scan-build: the valkey backend TUs were
"No bugs found" (Phase 2a sweep). Fuzz: `fuzz_respwire` (2.3M execs) +
`fuzz_valkey_dsn` (4.7M) no crash; both wired into the CI fuzz job (60s each).

E2E against a **real server** (`redis-server`; CI also falls back to docker
valkey/redis):

- `tests/e2e_valkey.sh` — full surface (set/get incl. binary+NUL, del/exists,
  TTL, incr, CAS four outcomes, prefix scan, clear, caps, kv/cache namespace
  isolation) **and the borrow-copy guard**, in **both runtimes**: `PASS (2 runtimes)`.
- `tests/e2e_feature_valkey.sh` — builds an `EMBED_PLATFORM=1` base (asserts it is
  valkey-free), builds the archive, `hull build --with=valkey` an app, runs it
  **under `--no-sandbox` and the real kernel sandbox** (network granted), asserts
  a plain app composes no valkey, and that the host allowlist denies a disallowed
  host before dialing: `PASS`.

`make check-feature-registry`: OK (installable + FEATURE_SPECS both include valkey).

## CI / release

New `valkey` CI job (installs redis-server, builds `HL_ENABLE_VALKEY=1`, runs
`e2e-valkey` then `e2e-feature-valkey`); `build-feature-valkey` release matrix
(linux-x86_64 / linux-aarch64 / darwin-arm64) publishing the signed
`libhull_feature-valkey-<arch>.a` (`TRUST_FEATURE_LIBS` chain). `docs/kv_cache.md`
corrected to describe the new `HlKvBackend` seam (the design draft mispredicted
`hl_db_feature_backends`).

## Known local build-directory incompatibility (pre-existing, not this feature)

`hull build --with=<feature>` fails to link against an ASan/UBSan (`make debug`)
`build/` tree — the composed platform lib carries sanitizer refs the plain
system-cc app link can't resolve. This is feature-agnostic (reproduces on
duckdb/postgres/mysql/gpu equally) and does not affect CI/release (they build
non-instrumented `EMBED_PLATFORM=1` trees). Workaround: `make clean && make
EMBED_PLATFORM=1` before composing. Filed as a separate build-system follow-up:
#295.

Please let the full CI matrix — MSan, static analysis, feature composition,
release-artifact construction, and both runtime jobs — finish before merge.
Auto-merge intentionally not enabled.


farkasmark:
## Merge-ready ✅ (not merging)

Full CI matrix is green on the tip (`22faec2d`): **47 SUCCESS + 1 SKIPPED** (the
skipped one is the optional Benchmark), run conclusion `success`. PR state:
`MERGEABLE` / `CLEAN`, auto-merge **not** enabled.

Heavyweight + Valkey-relevant jobs all passed: ASan+UBSan, **MSan+UBSan**,
**Static Analysis** (scan-build + cppcheck), Fuzz, both runtime jobs (Linux gcc +
Linux clang), Cosmopolitan/APE, the reproducible-build set, the release-artifact
feature builds (postgres/mysql/valkey/duckdb/gpu/tui), and the **Valkey/Redis
e2e** (compiled-in both runtimes + `--with=valkey` compose).

### Two CI failures found and fixed during review (both Valkey-scoped)
1. **cppcheck** — `cap/kv_feature.c` NULL-deref of the weak-hook result (guard moved
   into the loop condition, mirroring `db_select.c`, + a narrowly-scoped
   suppression) and a `cap/valkey_conn.c` pointer-subtraction warning (rewritten as
   `pairlen + 1`).
2. **Valkey e2e** — the readiness probe used `/dev/tcp` (a bashism); CI `sh` is
   dash. Replaced with a POSIX `redis-cli` PING in both e2e scripts.

### Final-review verification
- **cppcheck suppression is narrow.** A single `/* cppcheck-suppress
  knownConditionTrueFalse */` on the line directly above the one `if (bs && bs[i]
  && …)` statement. With `--inline-suppr` it applies to only that next line and
  only that check-id — it does **not** silence `nullPointer` (removing the guard
  would still fail cppcheck), and it is not a file-wide suppression.
- **The NULL guard is independently unit-covered.** New `test_kv_select` (4 cases,
  run under the green ASan/MSan/Linux jobs) drives every guard branch via a
  test-controlled strong override of `hl_kv_feature_backends`: `bs==NULL,n==0`
  (weak-default) → NULL; `bs==NULL,n>0` (malformed override) → NULL, proving `bs
  &&` fails closed without a deref; a NULL table entry → skipped; scheme
  match/miss/lowercasing + NULL/empty/schemeless DSNs. This is in addition to the
  e2e (base install-hint = NULL path; compiled-in/composed = match path).

Scope kept tight — only Valkey regressions were touched; the pre-existing
instrumented-tree compose limitation remains filed separately as #295, and
CacheLib is not started.

Ready to merge at your discretion.

PR #300, opened 2026-08-09
docs: design + benchmark plan for an in-stdlib O(1) LRU in _memstore (plan only, implementation unauthorized)

farkasmark:
## Summary

Docs-only. A design + benchmark **plan** for replacing `_memstore`'s O(n) LRU victim scan with an in-stdlib O(1) intrusive doubly-linked LRU. This is the sanctioned follow-up from the concluded native cache-store experiment (`docs/kvmem_negative_result.md`): fix the one workload that experiment improved (eviction-heavy) at its true root, inside the interpreter, with no scripting-boundary copy and no memory penalty.

**Plan only. No implementation exists.** Implementation remains unauthorized until this plan's semantic-parity and memory-impact review is approved. All validation in the doc is described as what the implementation PR *will* run, not as collected evidence.

## The change (proposed)

`make_room` today does a full `pairs(self.data)` min-`seq` scan per eviction → O(n) per op under sustained over-cap inserts. The plan threads a doubly-linked recency list through the **existing** entry records (`head`/`tail` + `prev`/`next`; the `seq` counter is removed) so eviction is an O(1) tail unlink. List maintenance is **gated on `evict=true`**, so the `hull/kv` (`evict=false`) get/set path does strictly less than today and cannot regress; the list exists only for `hull/cache`.

Scope: `stdlib/{lua,js}/hull/kv/_memstore.{lua,js}` only. No C, no bindings, no capability/manifest/gate change; the pure `_memstore` stays the sole memory backend.

## Covers the seven required review topics

Lua/QuickJS semantic parity · shared-namespace behavior · TTL + overwrite ordering · memory overhead · mixed-workload regression risk · eviction-heavy improvement · RSS measurement.

## Numeric ship gates (all required, per runtime)

1. Semantic differential (golden trace vs a frozen copy of today's `_memstore`): **zero mismatches**, both runtimes.
2. Mixed get/set (80/20 and 50/50), both policies: **≤ 10% regression** per runtime.
3. TTL-churn, both policies: **≤ 10% regression** per runtime.
4. Eviction-heavy: **≥ 5×** improvement per runtime (O(n)→O(1) floor).
5. Peak RSS (1e6 small entries): **≤ 10% above today** per runtime and policy.
6. Existing kv/cache unit + e2e tests: **zero regressions** (no edits).
7. No new C, dependency, or gate.

## List-integrity review checklist

Explicitly covers unlink on lazy TTL expiry / `del` / `clear`, repeated multi-victim eviction in one `make_room`, head/tail singleton transitions (empty↔singleton↔two-node, move-to-head of head/tail), shared-handle touches on one `Store`, and no-retained-reference-after-drop.

## Review gate

Reviewers: please assess the semantic-parity argument (§3), the memory-overhead claim (§6), and the numeric gates (§7). Implementation does not begin until this plan is approved.

### DIFF
--- a/docs/memstore_lru_plan.md
+++ b/docs/memstore_lru_plan.md
@@ -1,0 +1,436 @@
+# In-stdlib O(1) LRU for `_memstore` — design + benchmark plan (no implementation)
+
+**Status:** PLAN ONLY. No code until this plan's semantic-parity and
+memory-impact review is approved. This is the sanctioned follow-up from the
+concluded native-cache-store experiment
+([kvmem_negative_result.md](kvmem_negative_result.md)): fix the ONE workload that
+experiment improved (eviction-heavy, 30-90×) at its true root — an O(n) victim
+scan that lives entirely inside the interpreter — WITHOUT paying the
+scripting-boundary copy or the memory overhead that sank kvmem.
+
+Scope: `stdlib/lua/hull/kv/_memstore.lua` and `stdlib/js/hull/kv/_memstore.js`
+only. No C, no bindings, no capability change, no manifest/gate change. The
+`_memstore` is the semantic oracle both `hull/kv` (memory) and `hull/cache`
+(memory) depend on and that the retired kvmem was diffed against, so the bar is
+**observational identity with today's `_memstore`**, plus a strict eviction
+speedup and no memory regression.
+
+## 1. The defect (exact, from the current code)
+
+Both files are structurally identical. The victim search in `make_room`:
+
+```lua
+-- stdlib/lua/hull/kv/_memstore.lua:83-91  (JS: _memstore.js:55-64, identical)
+while over() and self.items > 0 do
+    local lru_k, lru_seq
+    for k, e in pairs(self.data) do                 -- O(items) FULL SCAN
+        if not lru_seq or e.seq < lru_seq then lru_k, lru_seq = k, e.seq end
+    end
+    if not lru_k then break end
+    drop(self, lru_k, self.data[lru_k])
+    self.st.evictions = self.st.evictions + 1
+end
+```
+
+Recency is tracked by a monotonic `seq` counter: `get` (hit), `put` (new or
+overwrite), and `incr` (existing) each do `self.seq = self.seq + 1; e.seq =
+self.seq`. The LRU victim is the entry with the lowest `seq`. Finding it is an
+O(items) scan, and under sustained over-cap inserts eviction fires on nearly
+every insert, so the store degrades to O(n) per operation. This is the entire
+performance gap; nothing else in `_memstore` is worse than O(1) amortized.
+
+Everything ELSE `_memstore` does (values, TTL `exp`, byte/item accounting,
+`stats`, `scan` order, error codes) is already O(1) per op and is **out of scope
+to change**.
+
+## 2. The fix: an intrusive doubly-linked LRU threaded through the entries
+
+Keep a doubly-linked recency list whose nodes ARE the existing entry records (no
+separate node objects, so no new per-entry allocation). The store gains `head`
+(most-recently-used) and `tail` (least-recently-used) references. Each entry
+record gains `prev`, `next`, and a self `k` (its own key, so the tail node can
+delete itself from `data`). The `seq` field and the counter are **removed** —
+list position replaces them.
+
+- **touch (move-to-head):** on `get`-hit / `put` (new or overwrite) / `incr`
+  (existing) — unlink the node from its current position and splice at head. O(1).
+- **evict (unlink-tail):** `make_room` drops `self.tail` repeatedly until within
+  caps. O(1) per eviction; total eviction work is O(evicted), not O(items).
+- **drop / lazy-expiry / del / clear:** every existing `drop` also unlinks the
+  node; `clear` resets `head=tail=nil`. O(1).
+
+Per-entry record shape (conceptual fields; the PHYSICAL layout is
+runtime-specific and chosen for memory in §6 — Lua uses positional/array-part
+records, not named fields):
+
+| today | after |
+|-------|-------|
+| `v, exp, bytes, seq` | `v, exp, bytes, k, prev, next` |
+
+Net: remove one integer (`seq`), add one string ref (`k`) and two node refs
+(`prev`,`next`). No new heap objects per entry. But field COUNT and per-runtime
+storage are not free — see §6, where the naive named-field layout is shown to
+fail the Lua RSS gate and a positional layout is required.
+
+**Gate the list on `evict`.** A `hull/kv` memory store has `evict=false` and
+NEVER evicts, so recency tracking is pure waste there. When `evict=false`, skip
+ALL list maintenance AND the recency bump entirely: `get` becomes `live + stats`
+(strictly less work than today's unconditional `seq` bump). The list exists only
+for `evict=true` (cache) stores, exactly where it pays off. This is what keeps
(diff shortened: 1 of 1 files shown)


farkasmark (commented):
## Formal review — validated against both `_memstore` implementations

**Verdict: NOT merge-ready as written.** The core algorithm (intrusive O(1) LRU, gate on `evict`) is sound and the semantic-parity argument holds, but four corrections are required — one critical (the Lua RSS claim is empirically false as specified). Implementation stays unauthorized.

### 1. Recency-touch points & operation ordering — ACCURATE ✅
Cross-checked every method in both files. Touch set is exactly `{get-hit, put-new, put-overwrite, incr-existing}`; `cas`-that-writes and `incr`-fresh inherit touch via `put`. Non-touching: `has`, `scan`, `cleanup`, `get`-miss, `get`-on-expired (`live`→`drop`, returns before the bump). Ordering is correctly distinguished: overwrite/incr promote-**then**-`make_room` (protects the written key); new-put `make_room`-**then**-link (entry not yet in the list). No correction needed; optionally state the `cas`/`incr`-fresh routing explicitly.

### 2. Removal / reset paths — CORRECT, one checklist gap ⚠️
`drop` is the single removal choke point for four of five paths — `live` (expiry), `del`, `make_room` (eviction), **and `cleanup` (eager sweep)** — so unlinking inside `drop` covers them all; `clear` is the only path that bypasses `drop` and must separately reset `head=tail=nil`. The plan calls out expiry/del/clear but **omits `cleanup()` from the §8 checklist**. Required: add `cleanup()` (a multi-drop sweep of scattered middle nodes) as an explicit checklist item.

### 3. Per-entry fields vs the 10% RSS gate — CRITICAL, plan claim is false for Lua ❌
Measured (Lua, `collectgarbage("count")`, 40k entries):

| layout | bytes/entry | vs today |
|---|---|---|
| named ×4 `{v,exp,bytes,seq}` (today) | 224 | — |
| named ×6 `{+k,prev,next −seq}` (as the plan specifies) | 320 | **+43%** |
| array ×6 `[v,exp,bytes,k,prev,next]` | 178 | −21% |
| hybrid 3-named + 3-array | 194 | −14% |

The plan's "net +2 slots, RSS ≈ flat" is wrong for Lua: table hash parts round to a power of two, so 4→6 named fields is 4→8 nodes (+43% entry memory) and **fails the 10% RSS gate outright**. Required corrections: (a) acknowledge Lua's power-of-2 hash sizing and QuickJS shape/inline-property storage explicitly; (b) mandate an **array-indexed (positional) entry record** as the Lua layout (leaner than today, −21%); (c) require the per-runtime RSS measurement to PIN the layout before implementation — QuickJS (+2 props is linear, lower-risk) still needs its own empirical check, not an assumption.

### 4. Benchmark isolation / warm-up / repetition / variance / baseline — UNDER-SPECIFIED ❌
§7 lists workloads and sizes but no rigor. Required additions: discard warm-up iterations; ≥5 (prefer ≥10) repetitions per cell; report **median + spread** (min/max or stddev), never a single number; `collectgarbage("collect")` before each timed span (Lua) and note GC pauses inflate variance; pin the baseline to a specific `_memstore_ref` commit + record the hull build config. Critically: **detect and FAIL on instruction-limit truncation** — a run that hits the 100M cap mid-populate silently reports a partial-run number (this actually happened in the retired kvmem RSS run, where the pure-Lua baseline appears to have died early, undercounting its RSS). A truncated run must error, not report.

### 5. Harness fairness — must be de-native-ized ❌
The preserved `tests/fixtures/kvmem_bench/` retains native-backend assumptions: the `nativemem.available` guard, `nativemem.get`, and (JS) the `_nativemem.js` ArrayBuffer `toBuf`/`fromBuf` conversion. Reusing it old-vs-new must: strip all native paths and the JS ArrayBuffer conversion (both sides are pure interpreter byte-strings — keeping the conversion would measure a boundary tax that does not exist for `_memstore`); substitute frozen `_memstore_ref` vs new `_memstore`; and use **equal op counts** old-vs-new (the native era's asymmetric sizing — native got 5× the ops because it was faster and to fit the budget — is inappropriate when both stores are the same speed class). The plan says "reuse and extend" without stating any of this; make it explicit.

### Not blocking
- Semantic differential, shared-namespace, TTL/overwrite ordering: validated, no correction.
- The `evict=false` skip-the-list optimization is observationally safe (no `stats` field exposes `seq`; `make_room` raises before evicting, so no list is needed).

**Required before implementation authorization:** corrections 2, 3, 4, 5 folded into the plan (3 is the gating one). Re-review the revised plan; do not implement until approved.

farkasmark (commented):
## Second formal review — revised plan (HEAD 2288fc7f)

The four required corrections from the first review are **incorporated and adequate**. One new methodology gap surfaced in the revision (below); it is a one-line clarification, not a design flaw.

### Corrections verified
1. **`cleanup()` in the checklist ✅** — §8 now has an explicit "unlink on `cleanup()` (eager sweep)" item calling out the scattered many-at-once removal (head/tail/middle/adjacent mix) as a distinct failure surface. Adequate.
2. **Positional Lua entries + readability ✅** — §6.1 mandates array-part records with the measured justification (named ×6 = +43% via Lua's power-of-two hash-node jump; positional = −21%), REQUIRES named index constants (`E_V…E_NEXT`) plus `lru_unlink`/`lru_push_head`/`lru_touch` helpers, and preserves the old field semantics in comments. The `exp = false` (not `nil`) note correctly avoids mid-array truncation. §6.2 states the QuickJS shape/inline-property assumptions explicitly (shared insertion order, `false` not `undefined`, positional-array alternative decided by measurement) with its own layout decision. Adequate.
3. **Separate measured RSS gates ✅** — §7.3 + ship-gate 5 gate RSS independently for each of four cells {Lua, QuickJS} × {evict=false, evict=true}, never blended, each with the truncation guard. Adequate.
4. **Benchmark rigor ✅** — §7.2 adds 2 discarded warm-ups, 10 timed reps/cell, forced GC before timing, median + spread reporting, §7.4 pinned baseline commit, and a HARD failure on instruction-limit truncation or incomplete op/entry counts (with the retired-kvmem partial-run cited as motivation). Adequate.
5. **De-native-ized harness ✅** — §7.1 requires stripping `nativemem`/availability guard and the JS `ArrayBuffer` conversion, frozen `_memstore_ref` vs candidate, identical inputs AND identical op counts (dropping the native-era asymmetric sizing). Adequate.

### One new clarification the revision introduced (minor, required before implementation)
§7.2 now states BOTH "one workload per process" (to fit the 100M instruction limit) AND "10 timed repetitions per cell" + "discard the first 2 iterations". Twelve executions cannot share one process at op counts large enough to time meaningfully — that would blow the instruction budget (the exact failure the §7.2 truncation guard now catches). The plan must make explicit that **each repetition is a separate `hull` process invocation**, aggregated by the harness driver (run the cell 12×, discard 2, median-of-10), rather than a 12-iteration in-process loop. "Iterations" currently reads ambiguously against "one workload per process". One sentence resolves it.

### Everything else
Semantic-parity argument (§3), touch-point/ordering analysis, shared-namespace (§4), TTL/overwrite ordering, and the `evict=false` skip-the-list optimization remain validated. §2's conceptual record shape now forward-references §6 for the physical layout, so they no longer read as contradictory.

### Merge readiness
The plan is sound and, as a **plan**, essentially ready. The single remaining item is the §7.2 repetition-structure sentence. Recommend folding it in (it falls squarely within the already-authorized benchmark-rigor correction); with that, the plan is complete for review sign-off. Implementation remains unauthorized pending final approval of the plan.

PR #301, opened 2026-08-09
docs(memstore): amend O(1) LRU plan — lazily-activated exact LRU (plan only, gates unchanged)

farkasmark:
## Summary

Docs-only amendment (new §10) to the approved `_memstore` O(1) LRU plan. The eager design was implemented and measured; it **passed** the semantic differential (0 mismatches), the eviction gate (Lua 11.7×, JS 26.4×), and **every RSS cell (0.99–1.00×)**, but **failed Gates 2/3** on the `evict=true` cache path (mixed Lua +60%/+85%, JS +27%/+38%; ttl Lua +49%, JS +31%). Root cause is structural: an exact O(1) LRU does a ~6-write move-to-head on every touch vs the old cheap `seq` bump, which is pure overhead on a cache whose working set fits.

**§10 proposes a lazily-activated exact LRU** that directly targets that failure. **No numeric gate changes** — §7 stands verbatim.

## The design (plan only, no code)

- **SEQ mode (default):** exactly today's cheap behavior — a `seq` counter bumped on touch, no links. `evict=false` stores never leave it; a cache with headroom stays here too.
- **First real eviction → activate:** reconstruct the exact LRU order from the unique `seq` values and switch the store permanently to intrusive-list mode (O(1) evictions thereafter).
- **No large temporary:** an intrusive stable merge sort over the existing link fields (slots reused; the 6-slot record is unchanged, so RSS stays flat).
- **clear()/empty → back to inactive.** Pre-activation `del`/expiry are seq-only; post-activation they unlink.
- **Shared handles** observe one store-level activation state.
- **Atomic activation:** a failure leaves the store and the triggering write unchanged (build writes only the unused slot before an allocation-free commit).
- **Rollover handling:** force activation early at a safe `seq` threshold so reconstruction always has unique seqs.
- **Added differential coverage:** activation boundary, multi-victim first eviction, expiry immediately before activation, counter rollover, clear/reactivation.

## Why this should work

Fitting-cache mixed/TTL keep the cheap SEQ path (never activate) → removes the +27%..+85% regression; eviction-heavy pays a one-time O(n log n) ordering cost then gets O(1) eviction → keeps the ≥5× win. All predictions are marked to-be-MEASURED, not assumed.

## Discipline

Every §7 numeric gate is unchanged and must pass as written. Stop condition: if the lazy design still fails any gate, the experiment is archived as a second negative result — no third design. `feat/memstore-o1-lru` (the eager implementation + full matrix) is preserved locally and **not modified** until this amendment is approved.

Reviewers: please assess §10.2 (reconstruction correctness), §10.3 (no-large-temp merge sort / RSS), §10.6 (rollover), and §10.8 (atomicity).

### DIFF
--- a/docs/memstore_lru_plan.md
+++ b/docs/memstore_lru_plan.md
@@ -178,6 +178,12 @@ risk is a slowdown on get/set-heavy traffic. Controls:
 
 Both must be benchmarked per runtime (Lua and QuickJS), never blended.
 
+> **Measured outcome + fix.** The eager design DID regress the `evict=true`
+> get/set path beyond the margin (mixed Lua +60%/+85%, JS +27%/+38%; ttl Lua
+> +49%, JS +31%). §10 amends the design to a **lazily-activated exact LRU** that
+> keeps the cheap `seq` path until an eviction actually occurs, directly removing
+> this overhead from the fitting-cache case without changing any numeric gate.
+
 ## 6. Memory overhead
 
 The list is threaded through the existing entry records (`prev`/`next` are
@@ -344,7 +350,9 @@ pinned baseline so a later reader can reproduce the exact comparison.
 
 **Ship gates for the eventual implementation review (numeric; all required).**
 Measured on arm64 Darwin, the redesigned `_memstore` vs today's `_memstore`,
-each runtime measured separately and never blended:
+each runtime measured separately and never blended. **These gates are UNCHANGED
+by the §10 amendment** — the lazy-activation design must clear every one of them
+exactly as written:
 
 1. **Semantic differential — ZERO mismatches.** The golden-trace differential
    (§3) reports 0 observable differences vs the frozen reference across the full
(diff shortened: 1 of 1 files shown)


farkasmark (commented):
## Formal review of the §10 lazy-activation amendment — focused on the transition

**Verdict: NOT approvable as written.** Two of the eight points expose critical, evidence-backed problems that couple together and may be design-invalidating; the other six need corrections/clarifications. Details below, with probe evidence (throwaway scripts, no _memstore change).

### Points 1 + 2 — six-slot layout is NOT interruption-safe (CRITICAL)

Hull enforces a per-request instruction budget via `lua_sethook(MASKCOUNT)` / `JS_SetInterruptHandler`; **stdlib code cannot disable that hook.** When the budget is exhausted the interpreter RAISES mid-execution. Activation is O(n log n) work inside one operation, so for a large-enough store it is interrupted partway.

The amendment's §10.8 atomicity claim does not hold on 6 slots. Post-activation needs BOTH `prev` and `next` (O(1) middle-unlink for move-to-head), i.e. slots 5+6. Pre-activation needs `seq` in slot 5. `commit` therefore overwrites slot 5 (seq → prev) entry-by-entry. A budget-kill mid-commit leaves some entries with a link in the seq slot and others with a number, `active` still false → the store is nominally SEQ mode but a later activation reads a table/false as "seq" and the merge-sort comparison crashes (`attempt to compare two table values`) or diverges. That violates "an interruption must leave valid SEQ or valid LIST state."

- **A 7th, dedicated `seq` slot fixes the atomicity** (activation writes only slots 6/7, never slot 5; the single `active=true` flip is the atomic switch; interrupt before flip → valid SEQ with seq intact, links recomputed next time; after flip → valid LIST). **But it costs ~16 B/entry.** The measured 6-slot store is already AT parity (matrix: Lua 96 MB both). A micro-probe here is confounded (it allocates a duplicate key string the 4-named layout lacks), so the 7-slot RSS must be MEASURED on the real 1e6 store — it is a genuine risk of breaching the 10% RSS gate that §6 currently claims is safe.
- A 6-slot journaled/resumable commit is possible but complex and does not help the large-store case below.

**Required:** prove interruption-safety explicitly with a per-phase slot table; adopt the 7-slot layout (or a journaled scheme) and MEASURE its real RSS against the gate.

### Points 4 + 5 — activation latency / instruction budget (CRITICAL, likely design-limiting)

Allocation-free bottom-up merge-sort probe (Lua, arm64 Darwin), which is the activation core:

| store size at first eviction | activation latency | ~instructions (est.) |
|---|---|---|
| 10,000 | 22 ms | ~4M |
| 100,000 | **331 ms** | **~66M (of the 100M default)** |
| 1,000,000 | did not complete (hit the Lua memory ceiling building the store) | ~660M → **exceeds the budget** |

Two consequences the amendment does not address:
1. **A large cache's first eviction cannot complete within the default per-request budget.** Above ~100k-150k entries activation alone approaches/exceeds 100M instructions, so the triggering op is killed, leaves valid SEQ state (with the 7-slot fix), and the NEXT eviction retries and is killed again — the store can **never** evict. The cache is effectively broken above a size ceiling.
2. **Even where it completes, the first eviction is a single-op stall of hundreds of ms** (100k → 331 ms of frozen event loop) that aggregate throughput hides.

**Required:** add the activation-latency benchmark and report the worst single-op latency (per the point); establish and DOCUMENT the maximum store size for which first-eviction activation fits the default budget; define behavior above it. If that ceiling is small relative to real cache sizes, this is the stop-condition trigger.

### Point 3 — sequence exactness / rollover

`SEQ_SAFE_MAX` must be a single constant **< 2^53** shared by both runtimes (JS `Number` is exact for integers below 2^53; Lua 5.4 int wraps at 2^63, so with the same sub-2^53 threshold Lua rollover cannot occur first). Also clarify: only `evict=true` stores ever consult `seq` (at activation), so an `evict=false` store's unbounded `seq` needs no threshold and its precision loss is harmless. **Required:** state the numeric threshold and the evict=false carve-out.

### Point 6 — expiry before reconstruction (the amendment is currently WRONG for parity)

The oracle's `make_room` scans ALL of `data` for min-`seq` and does NOT skip expired-but-unswept entries, so they ARE eviction candidates. Therefore activation must **retain** expired-unswept entries in the reconstruction; pre-sweeping them would DIVERGE from the oracle and fail the differential. Only entries the triggering op itself lazily dropped via `live()` (before `make_room`) are excluded — and that drop is identical in the oracle, so it does not affect the write's all-or-nothing insertion. **Required:** correct §10.9 to say expired-unswept entries are INCLUDED (not "removed before reconstruction"); distinguish the triggering-op lazy-drop (excluded, matches oracle) from unswept-expired (retained).

### Point 7 — merge-sort correctness cases

Activation never runs on an empty store (the `items>0` guard), but the sort must no-op safely on 0/1 nodes. **Required:** add pure-function merge-sort tests for empty, singleton, already-sorted, reverse-sorted, and large inputs; the probe confirms correctness at 10k/100k (sorted=true) but the large case is bounded by point 4.

### Point 8 — drain-to-empty via all paths

Deactivation-on-drain must live in the single `drop` choke point so del / lazy-expiry / cleanup / multi-victim-eviction all reset to SEQ mode by construction (all four route through `drop`). **Required:** specify the `items==0 -> deactivate` reset inside `drop`, and add differential cases draining to empty via EACH path then refilling to force reactivation.

### Bottom line

Points 3, 6, 7, 8 are fixable clarifications. Points 1+2 and 4+5 are coupled and serious: exact reconstruction is O(n log n) on one gas-limited, non-atomic operation, so the lazy design trades the eager design's per-touch overhead for a first-eviction latency/budget cliff and an interruption-atomicity gap that 6 slots cannot close. Before this amendment can be approved it must resolve the interruption-safe layout (with measured RSS) AND bound activation to the instruction budget with a documented size ceiling. If that ceiling is too low to be useful, the plan's own stop-condition applies: archive rather than attempt a third design.

farkasmark:
Closing unmerged. The approved stop condition is met: the eager exact-LRU design failed the mixed/TTL cache gate (+27-85%), and the lazy-activation amendment reviewed here is rejected (interruption-unsafe 6-slot transition, probable RSS regression at 7 slots, and an activation latency / instruction-budget ceiling that makes large caches un-evictable). Per the plan's own rule, no third exact-LRU design is attempted. Closure recorded in a docs-only PR; the O(1) LRU experiment is CONCLUDED — NOT SHIPPED, and _memstore stays unchanged.

PR #313, opened 2026-08-12
wasm: mapped-spans checkpoint 3a — public spans={} API + metadata (A–E)

farkasmark:
Checkpoint 3a: the public per-invocation mapped-span API on top of the merged
checkpoint-2 internal lifecycle (#309) + WAMR patch 0004. **A–E complete** — a WASM
plugin can now attach read-only file windows via `spans={}` and locate + read them
through the `SPAN_INFO` metadata query. Plan + locked decisions:
`docs/wasm_mapped_spans_checkpoint3.md` (docs PR #312).

### Items (all done)
- [x] **A** — windowed `fs.mmap({offset,length})` binding (Lua + JS); `e2e-spans-mmap`, 14 cases.
- [x] **B** — `HlWasmSpan.name[64]` + validated `hl_wasm_span_set_add`; `wasm_spans.name_validation`.
- [x] **C** — `HlWasmCallOpts.spans` + `HlWasmSpanReq`; shared Lua/JS parse+validate across all four call sites; `e2e-spans-bind`, 20 cases.
- [x] **D** — C-layer consumption + async ownership.
  - [x] D.1 malformed guard · D.2 pooled sync + D1-atomic + teardown-every-exit (7 tests, tsan-shared-heap) · D.3 persistent `inst:call` (5 tests) · D.4 async submission-pin ownership (deep-copy names + pin buffers into `HlWorkerWasmOp`, release in `op_free`; 4 C op tests + `e2e-spans-async` pooled+persistent, both runtimes; `tsan-worker-pool`).
- [x] **E** — `HL_WASM_OP_SPAN_INFO` metadata host-call; wire ABI `HlSpanMetaV1` (96B LE, decode-by-offset). Driven from a real WASM guest reworked into a **parameterised driver** (`spanprobe.wasm`: input picks the advertised cbSize, span index, destination mode, and a scratch sentinel; output returns the count query, index rc, the full 128-byte record, and the window's first byte read through the reported `base`). Full wasm32 matrix:
  - `e_span_info_record` — full record decode (version/struct_size/RO flag/name/base/len/foffset) + read the window via `base`.
  - `e_span_info_capacity_matrix` — caps {4,8,72,95,96,128}: exact `min(cap,96)` bytes written vs the 96-byte reference + untouched sentinel suffix; rc always 96.
  - `e_span_info_malformed_cap` — {0,1,2,3,4097,8192,65535} → -1, no record written, count query still 1.
  - `e_span_info_bad_dest` — high (0x7ffffff0) / would-wrap (0x80000000) → -1 without trapping the guest.
  - `e_span_info_boundary` — destination ending exactly at the linear-memory boundary → 96; one past → -1.
  - `e_span_info_out_of_range` — {1,2,99,-2} → 0.
  - `e_span_info_no_set_and_post_detach` — plain call → 0; spans call → 1; next plain call on the same cache → 0 (`tl_host_ctx.spans` cleared, distinct from never-had-spans).
  - `e_span_info_concurrent_isolation` — N threads each attach a span at a distinct file offset and see only their own `foffset` (thread-local; TSan covers the race).
  - Precise OOB-exception clear: `clear_validate_oob_exception` clears **only** the "out of bounds" exception our own `validate_app_addr` produced; the native-call-entry invariant (WAMR unwinds on any trap and never dispatches a further import) is documented, so no pre-existing exception is ever erased.

### Metadata ABI ceiling
`SPAN_INFO`'s destination is always an **unsigned 32-bit app offset** — a fixed property of the `(i32,i32,i32)->i32` `host_call` ABI, not a host-side high-address rejection. An address at/above `UINT32_MAX` is unrepresentable in the operand, so a Memory64 guest must place its scratch below 4 GiB; rejecting/trapping before narrowing such a pointer is **3b's SDK responsibility** and belongs in 3b's Memory64 tests.

### Prerequisite fixed in this branch
- **#316 (closes on merge)** — persistent-instance async self-rejected via the busy guard. One busy-state owner; worker uses `*_async` entries bypassing the reject; sync rejection preserved; busy cleared exactly once on every path. `e2e-persistent-async`.

### Known pre-existing findings (out of scope; none block CI)
- **#318** — Memory64 dispatch is compiled out of Hull's cap layer (`WASM_ENABLE_MEMORY64` is defined only for the vendored WAMR objects, never for Hull TUs), so `cap/wasm.c`'s detection + 8-cell marshalling are `#if`-excluded and `mod->is_memory64` is always 0. A genuine **Memory64 AOT `SPAN_INFO`** leg (low scratch succeeds; 64-bit `base` read works; invalid unsigned-i32 dest → -1) therefore cannot run through `hl_cap_wasm_call` today and is deferred to #318 — it activates previously-dead dispatch paths and needs its own build/include change + validation pass, out of 3a scope. Docs (`wamr_architecture.md`, `CLAUDE.md`, `roadmap.md`, `stability.md`) corrected to stop claiming Memory64 works through the cap layer.
- **#314** gas-trap classification · **#315** segment shared-heap descriptor leak · **#317** Lua `compute.async.call` hangs on a worker trap (JS unaffected; the async gas-trap e2e case is JS-only, span trap-cleanup proven at C level).

### Follow-on — checkpoint 3b (separate PR)
SDK ergonomics header `hull/wasm/span.h` (typed accessors over `HlSpanMetaV1`), an example plugin, `hull compute refresh-header` wiring, the full interp+AOT e2e, the Memory64 AOT `SPAN_INFO` leg once #318 lands, and the native/differential test. Today the guest reads a span via the raw `host_call(SPAN_INFO)` (as `spanprobe.wasm` does).


farkasmark:
Final CI green on `74247ebb1f0d2abbd817d84d5bd5e8ea736c2a0c` (docs + SPAN_INFO ABI-comment tip): **47 pass, 1 skipping, 0 fail**.

Run: https://github.com/artalis-io/hull/actions/runs/31647394385

Spans-relevant legs all pass, including **TSan (WAMR shared-heap destroy + guarded subrange + spans)**, **TSan (worker pool)**, **Static Analysis**, **GPU feature build (Linux)** (the earlier red was a transient upstream `fetch-wgpu` SHA-256 blip — not retried, checksum untouched — and cleared on its own), and all **Reproducible build** legs. Full wasm32 `SPAN_INFO` E matrix passes locally under ASan.

Moving from draft to ready for review. The Memory64 AOT `SPAN_INFO` leg is deferred to #318 (Memory64 cap-layer dispatch is compiled out of the shipped build), as noted under known pre-existing limitations.

PR #339, opened 2026-08-16
bench: mapped-span performance benchmark (measures HullSpan vs native mmap)

farkasmark:
Closes the one unmet acceptance criterion of the mapped-spans arc: the "HullSpan read approaches native mmap" claim, now **measured** rather than asserted. Design/methodology locked in `docs/mapped_span_benchmark_design.md` (D1-D11), which was signed off before implementation.

## What it does

Four workloads (sequential bytes, sequential u32/u64, deterministic-LCG random access, parser-like record scan) as **one shared body** (`bench_span_ops.h`) compiled on two targets (native + wasm32) through the shipped `hull_span.h` accessors, so every implementation exercises the same inlineable, bounds-checked read path. Three implementations per workload:

- **native mmap** (the baseline/target) - always runs, any arch, no wamrc needed;
- **HullSpan AOT** (under test) - a host-mmap'd file window read by a wamrc-built AOT wasm32 guest via a HullSpan;
- **copy-once** - the whole dataset copied into wasm linear memory (the copy baseline); above the wasm32 256 MB I/O ceiling it is reported *"not representable within the configured linear-memory limit"*, not a failure.

## Methodology (the parts that matter)

- **Steady-state is not end-to-end call time renamed.** Two-point `(t(1), t(K))` amortization isolates per-scan cost from attach/dispatch/teardown: `steady = (t(K) - t(1))/(K-1)`. Raw end-to-end timings are retained in the JSON.
- **Correctness gate before any timing.** Every representable impl's checksum must equal native, else the workload fails.
- **Zero host calls in the scan loop, proven at runtime.** A new `hl_cap_wasm_host_call_count()` counter measures per-scan host-call delta; it must be 0 (`hull_span_setup`'s SPAN_INFO call runs once, before the rep loop). The static companion (G4) confirms via `wasm-objdump` that `host_call` appears only in the 2 setup sites - labelled accurately as WASM-bytecode inspection, **not** a claim about native AOT codegen.
- Medians + MAD dispersion over iterations; getrusage page-fault + RSS capture; deterministic dataset; reproducible JSON artifact.

## Gates (`tests/bench_mapped_span.sh`)

G1 drift (guest wasm reproduces from the canonical SDK header), G2 must-not-skip (every workload ran AOT+span), G3 in-process correctness + zero-host-call, G4 static bytecode inspection. **The <=10-15% steady-state overhead comparison is reported as an artifact, not gated**, until per-arch baselines stabilize (D11).

## CI

- Required PR job `mapped-span-bench` (96 MiB, x86_64) in `ci.yml`.
- Separate manually-triggered `bench_mapped_span_1g.yml` (`workflow_dispatch`, default 1 GiB) demonstrating the copy-once ceiling finding at scale.

## Local validation

The harness compiles under full production CFLAGS and links the real cap layer; the native measurement path produces deterministic checksums with sound two-point steady-state + MAD; the guest wasm compiles and `wasm-objdump` confirms `host_call` in exactly the 2 setup sites; the runner's G1 passes and G2 correctly fails without wamrc (proving the must-not-skip gate). **The AOT wasm rows + the overhead numbers are produced by this CI run** - wamrc is not buildable locally (vendored WAMR compiler vs local LLVM 22).

Scope is strictly the performance criterion: no fuzzer, no buffer-ownership refactor (both tracked separately).


farkasmark:
Both blocking issues fixed (merge stays paused per your instruction). Commit `4f1997ba`.

**1. Chunked-copy is now genuinely measured.** It processes the file through a fixed **bounded** linear-memory buffer, the host re-filling native→wasm per chunk (guest LINEAR mode) — so it's representable even far above the copy-once ceiling. It's checksum-decomposable and the correctness gate proves it equals native for the three streamable workloads: `seq_bytes` (any cut), `seq_words` (8-aligned cuts), `parser` (record-aligned cuts). `random` is reported `representable=-4` (not-applicable): a random walk can't be served from a chunk already discarded — the exact property a span provides and a bounded chunked copy cannot. That's a published finding, not a silent gap. Four impls → four JSON rows.

**2. The host-call counter costs production nothing.** The atomic + accessor are now `#ifdef HL_WASM_HOST_CALL_COUNTER`, compiled only into a bench-private `build/bench_cap_wasm.o`. The shared production `build/cap_wasm.o` carries neither the symbol nor the atomic — verified with `nm` (no `host_call_count`). The public header hides the declaration behind the same flag, so it's not a production API.

**Bonus fix surfaced while making the required CI job actually pass:** WAMR meters AOT against an `INT_MAX` (~2.1e9) instruction ceiling, so K internal reps of a 64–128 MiB scan blow it and the `t(K)` call fails with GAS at the mandated dataset size. Switched to the setup-only control D3 explicitly sanctioned: `steady = median(t(1 scan)) − median(t(0 scans))`, each call running ≤1 scan (gas-safe at any size). The zero-host-call proof is unchanged in spirit — `hostcalls(1 scan) == hostcalls(0 scans)`.

Also embedded the committed guest as an **interpreter fallback**, so the wasm impls + correctness gate run without wamrc; the JSON `engine` field is `aot|interp` and CI asserts `engine=aot`. Validated locally under the interpreter: all four impls produce checksums identical to native (0 failures), chunked correct for the 3 streamable workloads and n/a for random, zero host-call-in-scan. The AOT perf numbers come from this CI run.

(Re the commit-count note — you were right, it's now five+ commits; I won't quote a count.)

farkasmark:
Round 2 — all three addressed (merge stays paused). Commit `b4773d34`.

**1. Chunked random access is now measured, not `-4`.** A bounded reader serves the fixed-LCG offsets from a one-page (4 KiB) cache, (re)loading the page containing each scattered read. It's representable — checksum matches native by construction (verified: `282187804181994` == native at 8 MiB) — it just **thrashes**, and the row now reports the metrics that make that the *useful* comparison:

```
random  chunked_copy  representable=1
  chunk_loads=131096  cache_hits=393192  bytes_copied=536,969,216 (~512 MiB)
```

~512 MiB copied to read ~512 KiB of actual data — **~1000× amplification** vs a span reading 4 bytes in place. The random read count is capped (`BENCH_RANDOM_MAX_READS`, applied identically to native/span/chunked so checksums stay matched) so the reload volume is bounded rather than terabytes. The full workload × baseline matrix is filled; G2 now requires 4/4 representable chunked rows + nonzero chunked-random `bytes_copied`.

**2. Scoped to warm steady-state (your option b).** The `cache="cold"` path is removed — it faulted pages back in before the timed scan and only touched the native mapping, never the span, so it never measured cold. A real per-iteration cold protocol (evict *both* mappings each iteration, verified major-fault evidence) + RSS validation is tracked as a follow-up, not claimed here. Removed from the harness, the `CACHE` env, and the 1 GiB job input; the design record's cache-state section says so explicitly.

**3. G1 now hard-checks provenance.** New **G1a** asserts the committed guest `.wasm` sha256 equals the value in `bench_span_guest.wasm.prov` (fails if the binary changed without updating provenance). The cross-toolchain byte-identity check stays a WARN (wasm codegen isn't reproducible across clang/lld); a non-compiling source still hard-fails.

Validated locally under the interpreter: all four impls incl chunked-random match native (0 failures), thrash metrics reported, G1a matches, no gas warnings. AOT perf numbers from this CI run.

PR #341, opened 2026-08-16
bench: clamp mapped-span bench ITERS/WARMUPS (c-audit low-sev UB fix)

farkasmark:
Follow-up from `/c-audit` of the merged mapped-spans benchmark + fuzzer/RSS work.

The audit found the codebase clean except for one genuine, low-severity, dev-only UB in the benchmark harness: `ITERS`/`WARMUPS` are parsed with bare `atoi()`, so `ITERS=0` makes `stats()` read `a[-1]`/`a[0]` of an empty array and a negative value sizes the timing mallocs from a negative int widened to a huge `size_t`. Clamps `g_iters >= 1`, `g_warmups >= 0`. Regression: `DATASET_MB=2 ITERS=0` now exits 0.

Everything else was clean — see the audit summary in the thread. The shipped production change (the `cap/wasm.c` host-call counter) is gated behind `HL_WASM_HOST_CALL_COUNTER` and verified absent from the production object; no JS/Lua logic was in scope.

### DIFF
--- a/bench/wasm/bench_mapped_span.c
+++ b/bench/wasm/bench_mapped_span.c
@@ -505,6 +505,11 @@ int main(int argc, char **argv)
     if ((e = getenv("ITERS")))   g_iters   = atoi(e);
     if ((e = getenv("WARMUPS"))) g_warmups = atoi(e);
     if ((e = getenv("OUT")))     g_out     = e;
+    /* Clamp to sane floors: a bad env value (ITERS=0/negative) would otherwise
+     * make stats() read a[-1]/a[0] of an empty array and size the timing mallocs
+     * from a negative int -> huge size_t. g_iters must be >= 1; warmups >= 0. */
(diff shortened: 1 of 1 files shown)


farkasmark:
Folded in L2 + L3 from the audit for consistency (commit `1a2f4b0a`):
- **L2**: all 11 `malloc()` sites now go through a fail-fast `xmalloc()` (OOM → clean diagnostic exit, not a silent deref); the guaranteed-non-NULL `input` let the copy-once read path drop its dead `!input` check.
- **L3**: `chunk_plan`'s growable-array realloc now uses `xrealloc()` (frees old block + exits on failure, no leak-on-fail) and `PUSH` rejects int overflow of the cap doubling before the realloc.

Regression: normal run 0 failures, `ITERS=0` exits 0, fuzzers still clean. This closes every audit finding (all were low-severity, dev-only benchmark robustness).

PR #343, opened 2026-08-17
feat(source): hull.source.lua design + slice 1 (lexer, ranges, line map)

farkasmark:
Foundation for Hull's future build-plugin/metaprogramming system: a small, reusable, **Hull-owned** Lua source-analysis API (AST + comments + annotations + exact source ranges + syntax diagnostics). No consumer (`hull analyze` / `hull.query` / `hull.compute` / codegen) is built here.

## Design decision (committed record: `docs/lua_source_analysis_design.md`)
LuaLS was the preferred starting point but **rejected**: its tokenizer needs LPegLabel (a native C lib — violates the no-native-dep / no-C-parser rules) and its parse is coupled to the semantic `guide`/`compile` layer, tripping the spec's stop conditions. Chosen: a **compact pure-Lua recursive-descent Lua 5.4 parser** (same pattern as `template.lua`), a replaceable detail behind the `hull.source.lua` contract. All five reviewer contract tightenings are locked in the design (full Lua 5.4 required; `parse()` returns `(unit, err)` with syntax→diagnostics/`err==nil`; resource limits; byte-based half-open ranges with on-demand line/col; Lua 5.4 differential conformance corpus in the harness).

## Slice 1 (this PR): lexer + ranges + line map
- `source/range.lua` — neutral `SourceRange` (`[start, stop)`, 1-based bytes) + line-start index, on-demand `(line, col)` (JS-reusable).
- `source/diagnostic.lua` — the shared structured Diagnostic.
- `source/lexer.lua` — complete Lua 5.4 lexer with exact half-open byte ranges; long brackets at any `=` level, every numeral (validated via `tonumber`, not `load()`) + escape form, shebang, CRLF/LF/no-final-newline. **Never raises** — malformed → diagnostic, always advances. Resource limits bound adversarial input.
- `source/lua.lua` — the public `parse()` contract (no raw error crosses; `unit:text/:position/:line_col`; `unit.ast` arrives in slice 3).

## Tests
`test_lexer.lua` (**83 cases**) via a vanilla `lua_State` harness (`tests/hull/source/test_lua_source.c`, links only vendored Lua 5.4, auto-discovered by `make test`): exact byte ranges incl. UTF-8-before-token, CRLF/LF/no-newline/shebang, long brackets at multiple levels, valid+malformed numerals/escapes, keyword-vs-identifier boundaries, every resource limit, lossless slicing via `unit:text()`, and no-raw-exception-crossing-`parse()`. luacheck clean. The `load()` differential corpus stays in the harness (later slice), never the module.

Next slices: expression grammar → statements/declarations → comments/annotations → recovery/conformance. Consumers wait until the slice 3–4 AST tests are stable.

### DIFF
--- a/docs/lua_source_analysis_design.md
+++ b/docs/lua_source_analysis_design.md
@@ -1,0 +1,251 @@
+# `hull.source.lua` — Lua source-analysis layer (design / decision record)
+
+**Status:** DESIGN, locked for implementation. Establishes the reusable,
+Hull-owned Lua source-analysis foundation (AST + comments + annotations + exact
+source ranges + structured diagnostics) that later powers `hull analyze`, generic
+build plugins, `hull.query`, `hull.compute`, derive/codegen, schema generation,
+and editor/agent tooling. **None of those consumers are built here.** This layer
+only parses/analyzes source structure (no type-checking, no bytecode/WASM, no IR,
+no source mutation, no module loading — §14).
+
+## 0. Decision: a Hull-owned pure-Lua parser, not vendored LuaLS
+
+The preferred starting point was LuaLS. Investigation (against LuaLS's actual
+architecture) trips the spec's stop conditions:
+
+- LuaLS's tokenizer depends on **LPegLabel**, a **native C library** — violates
+  "no native dependencies" and "no C parser" (keep parsing out of the trusted
+  core). The tool VM ships no extra native libs.
+- The syntactic parse is **coupled** to `parser/guide` + `parser/compile` (the
+  semantic phase this layer explicitly excludes); a syntax-only extraction is a
+  large, surgical subset.
+- LuaLS node positions are an internal encoding, not raw byte offsets; and its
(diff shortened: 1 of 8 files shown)


farkasmark:
Three lexer-contract fixes applied (commit `9a3bf819`):
1. **Escaped CRLF/LFCR** in short strings is now one escape sequence (`\<newline>` handled as LF/CR/CRLF/LFCR); tests assert zero diagnostics + lossless ranges for all four.
2. **Comments bounded** — added `max_comments`, lowered the `max_tokens` default (2M→500k) for a defensible memory bound, and routed limit diagnostics through a terminal path that bypasses `max_diagnostics` (so a full list or `max_diagnostics=0` can't suppress the stop reason).
3. **opts/limits validated at the boundary** — non-table opts/limits, non-string path, negative/non-integer limits return a clear `(nil, err)` before the lexer, not a `lua.internal` via the pcall.
Plus: `unit:position` clamps offsets outside `[1, #source+1]`. 110 tests pass (was 83); luacheck clean.

PR #344, opened 2026-08-17
feat(source): hull.source.lua slice 2 — expression grammar

farkasmark:
Slice 2 of the Hull-owned Lua source-analysis layer (design: `docs/lua_source_analysis_design.md`; slice 1 in #343). Builds the recursive-descent **expression** parser on the slice-1 token stream.

## What lands
- `source/parser.lua` — Lua 5.4 expression grammar with **exact half-open byte ranges** on every node and the complete precedence/associativity table (lparser.c priorities: right-assoc `..`/`^`, unary priority 12, `and`/`or`/`not` as keyword operators). Nodes: `literal` (number/string/nil/boolean), `vararg`, `name`, `field`, `index`, `call`, `method_call`, `table` (`field_item`/`field_name`/`field_expr`), `unary`, `binary`, `paren`, `function_expr`.
- **Never raises** — a parse error emits a `lua.syntax` diagnostic + an `{kind="error"}` node; nesting bounded by `max_depth` (no stack overflow). A `function` expression's body is a statement block (slice 3), so it's balanced-skipped to `end` with a `lua.unsupported` diagnostic for a non-empty body (valid syntax not yet represented → explicit, never a silent malform); `function() end` parses cleanly.
- Lexer fix surfaced by `#t`: a shebang is only `#!`, not a bare leading `#`, so the length operator lexes correctly.

## Tests
`test_parser.lua` (**91 cases**) via the existing vanilla-`lua_State` harness: literals, all suffix/call forms, chained suffixes, table constructors (every field kind + `,`/`;`), unary, and **precedence/associativity asserted on tree shape** (`1+2*3`, `2^2^3` right-assoc, `a..b..c` right-assoc, `1-2-3` left-assoc, `-2^2`, parens), `function_expr` params + deferred-body diagnostic, error diagnostics with no raises, `max_depth`. Lexer suite still **110/110**. luacheck clean.

Next: slice 3 (statements/declarations — completes `function_expr` bodies and the full `parse()` AST). Consumers still wait until slices 3–4 stabilize.

### DIFF
--- a/stdlib/cli/lua/hull/source/lexer.lua
+++ b/stdlib/cli/lua/hull/source/lexer.lua
@@ -110,8 +110,10 @@ function M.tokenize(source, opts)
     local function is_space(c) return c == " " or c == "\t" or c == "\r"
         or c == "\n" or c == "\v" or c == "\f" end
 
-    -- ── shebang: a leading '#' line (Lua's loader skips it) ───────────
-    if at(1) == "#" then
+    -- ── shebang: a leading '#!' line (Lua's file loader skips it). Only "#!",
+    -- NOT a bare '#', so a chunk/expression beginning with the length operator
+    -- (e.g. `#t`) lexes '#' as an operator rather than eating the line. ─────
+    if at(1) == "#" and at(2) == "!" then
         local e = source:find("\n", 1, true)
         local stop = e and (e + 1) or (n + 1)   -- include the newline if present
         comments[#comments + 1] =
--- a/stdlib/cli/lua/hull/source/parser.lua
+++ b/stdlib/cli/lua/hull/source/parser.lua
@@ -1,0 +1,339 @@
+--
+-- hull.source.parser — recursive-descent Lua 5.4 parser (slice 2: expressions).
+--
+-- Consumes the lexer token stream and builds Hull AST expression nodes with EXACT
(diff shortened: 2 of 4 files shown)


farkasmark:
Three review fixes applied (commit `34272641`):
1. **for/while skipper bug fixed** — for/while open exactly one end-terminated block; their `do` is absorbed via a `pending_do` counter (correct even with a nested function literal in a loop header); `repeat`/`until` isn't end-terminated. Tests cover nested if/for/while, standalone `do`, `repeat`, and nested functions — the reported false 'unterminated' is gone.
2. **Parser diagnostics respect the configured limit** — `max_diagnostics` carried into parser state; `P:err`/`P:unsupported` are capped, the `max_depth` diagnostic is terminal (always records), matching the lexer policy.
3. **Literal shapes locked** — every literal has `text` (exact lexeme); booleans additionally expose boolean `value`; `nil` has `subtype="nil"` and no `value`. Shape assertions added for all subtypes.

Also: `primary()` consumes the offending token on error (no recovery stalls). 131 parser tests (was 91); lexer 110/110; luacheck clean.

Workspace note: `deliverables/` (unrelated LaTeX/PDF overview docs) is intentionally left untouched and out of this PR.

PR #345, opened 2026-08-17
feat(source): hull.source.lua slice 3 — statements, declarations, full AST

farkasmark:
Slice 3 completes the Lua 5.4 parser: the statement layer, real function bodies (the slice-2 balanced-skip stub is gone), the full chunk AST from the public `parse()`, and the `walk`/`is` traversal helpers. Design: `docs/lua_source_analysis_design.md` (slice 1 #343, slice 2 #344).

## What lands
- **Statements**: `local_declaration` (with `<const>`/`<close>`), `assignment` (multi-target), `call_statement`, `do`, `while`, `repeat`, `if`/`elseif`/`else`, `numeric_for`, `generic_for`, `function_declaration` (global / dotted / method / `local function`, with `is_local`+`is_method`), `return`, `break`, `goto`, `label`. `exprstat` disambiguates assignment vs call the standard way.
- **Real function bodies** — `funcbody` parses the body block, so function expressions *and* declarations carry a statement `body`.
- **`parse()` produces `unit.ast`** = `{kind="chunk", body, range}`; lexer+parser diagnostics combined (each phase capped by `max_diagnostics`).
- **Bounded block nesting** (shared `depth` budget) → a terminal `lua.limit.max_depth` instead of a Lua stack overflow on deeply nested `do`/`if`.
- **Best-effort recovery** — stall guard in `block()` + token-consuming `primary()` keep the chunk parsing past a bad statement.
- **`lua.walk(ast, fn)`** (deterministic pre-order DFS in source order) and **`lua.is(node, kind)`**.

## Tests
`test_statements.lua` (**101 cases**): every statement kind, function bodies, control flow, break/goto/label/return, `walk`/`is` determinism, the **§41 acceptance example** (score function + query declaration discovered *structurally through the Hull API*), best-effort recovery, and a no-raise / bounded-nesting boundary. Slice-2 body assertions updated (a non-empty body now parses cleanly instead of emitting `lua.unsupported`). Totals: lexer 110, parser 133, statements 101. luacheck clean.

Next: slice 4 (comments/annotations — the generic `---@` scanner + declaration attachment, completing `lua.annotation`), then slice 5 (recovery + the Lua 5.4 differential conformance corpus).

### DIFF
--- a/stdlib/cli/lua/hull/source/lua.lua
+++ b/stdlib/cli/lua/hull/source/lua.lua
@@ -22,6 +22,7 @@
 --
 
 local lexer = require("hull.source.lexer")
+local parser = require("hull.source.parser")
 local range = require("hull.source.range")
 local diag = require("hull.source.diagnostic")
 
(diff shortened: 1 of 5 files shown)


farkasmark:
Four review fixes applied (commit `798c4c88`):
1. **Assignment targets validated** — each must end as name/field/index; a call/literal/paren *as* the target emits `lua.syntax` (a call *inside* a valid target, `f().x = 1`, stays valid); the assignment node is retained for recovery.
2. **Local attribute restricted to `<const>`/`<close>`** — text preserved for source fidelity, any other name diagnosed.
3. **`max_diagnostics` is now the authoritative SourceUnit bound** — lexer+parser diagnostics combined, every terminal `lua.limit.*` kept, ordinary capped at the total (was ~2×).
4. **`numeric_for.var` is now a `{name, range}` record**, consistent with generic-for/local names.
Plus the stale parser header is refreshed. Targeted node-shape tests added; lexer 110, parser 133, statements 130; luacheck clean.

PR #346, opened 2026-08-17
feat(source): hull.source.lua slice 4 — the ---@ annotation layer

farkasmark:
Slice 4 of `hull.source.lua`: the generic `---@` annotation layer, attaching to the stable, range-bearing declaration nodes from slice 3 without touching statement shapes or diagnostic budgeting.

## New module: `hull.source.annotations`
- **Scanner** `parse_comment` — a LINE comment `---@name`, `---@name(args)`, or `---@name text` → `{ name, args?, text?, raw, range }`. Whitelist-free: any name is captured (`---@derive`, `---@route`, `---@query`) with the same fidelity as LuaDoc tags. `--@` (two hyphens) and long comments are never annotations.
- **Attachment** `attach` — the annotation comments in the unbroken run of **leading** comment lines directly above a declaration attach to it. Blank line / code / trailing-comment-on-a-code-line breaks or is excluded. Nested declarations get their own runs. Node carries `annotation_list` (ordered, repeat tags like `@param` all kept) + `annotations` (name → first). Annotation comments in `unit.comments` are retagged `kind = "annotation"` per the contract, with the parsed record hung off the comment.

## Public surface (`lua.lua`)
- `lua.annotation(node, name)` and `unit:annotations_for(node)`.
- `attach` runs `pcall`-guarded after unit construction — an attachment bug degrades to "no annotations", never fails a clean parse.

## Tests
`test_annotations.lua` (59 cases): scanner shapes, non-annotations, contiguous / blank-broken / trailing / nested attachment, repeat-tag retention, comment retagging, range fidelity, the §41 acceptance example, and a never-raises boundary.

Suites: **lexer 110 + parser 133 + statements 130 + annotations 59** via the vanilla `lua_State` harness. luacheck clean; full `hull` build OK (module embeds into the binary). No `load()` / dynamic compilation.

### DIFF
--- a/docs/lua_source_analysis_design.md
+++ b/docs/lua_source_analysis_design.md
@@ -164,17 +164,26 @@ tracked explicitly (no unbounded native Lua recursion).
   annotation`, in `unit.comments` (source order).
 - **Annotations are Hull's own scanner over comment text — no known-tag
   whitelist.** Any `---@name`, optional `(args)`, optional trailing text becomes
-  `{ name, args = <string?>, raw = <full comment>, range }`. Arbitrary
-  `---@query`, `---@compute`, `---@derive(json)`, `---@some_future_plugin(a,b)`
-  all survive generically. Known LuaDoc tags (`@param/@return/@field/@class`)
-  MAY additionally expose parsed fields, but the generic form is always present.
+  `{ name, args = <string?>, text = <string?>, raw = <full comment>, range }`.
+  Arbitrary `---@query`, `---@compute`, `---@derive(json)`,
+  `---@some_future_plugin(a,b)` all survive generically. Known LuaDoc tags
+  (`@param/@return/@field/@class`) MAY additionally expose parsed fields, but the
+  generic form is always present.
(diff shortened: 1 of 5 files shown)


farkasmark:
Two contract fixes applied (commit `daa38b04`):
1. **Attachment failure no longer swallowed** — `lua.parse()` runs `attach` `pcall`-guarded and, on internal failure, appends an explicit `lua.internal` diagnostic to `unit.diagnostics` (so `#unit.diagnostics == 0` reflects the degraded state) instead of returning a clean unit with annotations silently missing. The valid AST is still returned.
2. **Scope locked to declarations only** — `local_declaration` + `function_declaration` (was: every statement kind). A `---@` run above any other statement attaches to the nearest following declaration instead, or to nothing. Design §8 updated to state declaration-scoping explicitly.
Tests +14: declaration-only scope, plain `--`/`--[[ ]]` comments participating in a contiguous leading block while blank/code/trailing break, and an injected attach failure surfacing as `lua.internal`. lexer 110 + parser 133 + statements 130 + annotations 73; luacheck clean.

PR #348, opened 2026-08-17
test(source): Lua source-analysis conformance harness (+ first lexer conformance fix)

farkasmark:
Step **A** of the post-build-plan roadmap: a differential conformance harness validating `hull.source.lua` against **real Lua 5.4** (the same vendored `load()` the test harness links — in-process, compile-only), plus the first real bug it caught.

## Design
`docs/lua_source_conformance_design.md` (committed here). Directional invariant: Hull is a *syntactic* recognizer, `load()` is syntactic + a thin static-semantic layer, so a Hull reject must imply a `load` reject (false-reject is the key bug class), while a Hull accept may diverge from a `load` reject only for a **pinned semantic case** (break/goto/vararg/const/attribute/limits), classified by an anchored, guarded allowlist of Lua-5.4 error fragments.

## Harness (`UTEST(lua_source, conformance)`)
- **Corpus in C**: `dirent` walk over `stdlib/lua`, `stdlib/cli/lua`, `examples`, `tests/fixtures` — regular `.lua` only, symlinks skipped (no traversal loops), sorted; exposes `{path, source}` records so oracle and parser get **identical bytes** (no `io`, no second FS pass).
- **`test_conformance.lua`** (its own leg for isolated failures): oracle-classified corpus (load-accepted ⇒ Hull clean; load-rejected ⇒ directional/semantic rules, reported separately), range round-trip over **AST child nodes only** (annotations excluded), curated syntactic negatives, pinned semantic divergences, **direct classifier tests** (each fragment matches its case and *not* nearby syntax errors), and a **fixed-seed mutation fuzz** (N=20/seed, 4000-mutant cap). `lua.limit.*`/`lua.internal` excluded from accept/reject; divergences reported by category + count.

Result over **224 corpus files**: 0 false-rejects, 0 range violations, **0 fuzz divergences**; 517 conformance assertions.

## First catch: a real lexer conformance bug
The fuzz found that Hull terminated a short comment **only at `\n`**, but real Lua 5.4 ends it at `\n` **or** `\r` (`llex.c` `currIsNewline`). A lone `\r` in a comment made Hull swallow post-CR code as comment and *accept* input Lua rejects — a fidelity bug for lone-CR / mixed-ending files. Fixed the short-comment scan and `range.linemap` (lone `\r` and `\r\n`/`\n\r`-as-one-break) to match Lua exactly; the shebang finder is correctly left `\n`-only (matches `luaL_loadfilex`). +5 regression tests (lexer 110 → 115).

Follow-up (documented, not in this PR): a continuous libFuzzer never-raise harness.

Suites: **lexer 115 + parser 133 + statements 130 + annotations 77 + conformance 517**; luacheck clean; full `hull` build OK.

### DIFF
--- a/docs/lua_source_conformance_design.md
+++ b/docs/lua_source_conformance_design.md
@@ -1,0 +1,241 @@
+# Lua source-analysis conformance harness (design)
+
+Status: DESIGN (pre-implementation). Companion to
+[lua_source_analysis_design.md](lua_source_analysis_design.md). This is step **A**
+of the post-build-plan roadmap: validate `hull.source.lua` against real Lua 5.4 as
+ground truth, systematically, before any consumer (`hull analyze`, `hull.query`,
+`hull.compute`, codegen) is built on the AST.
+
+## 1. Goal
+
+Turn "the tests I hand-wrote pass" into "the parser agrees with real Lua 5.4 across
+thousands of files, and its byte ranges are internally consistent." Concretely, catch:
+
+1. **False rejects** — the parser emits a syntax diagnostic on input real Lua
+   accepts. This is the highest-value bug class: real code the layer would choke on.
+2. **False accepts** — the parser reports clean on input real Lua rejects for a
+   *syntactic* reason (a missed syntax error).
+3. **Range-integrity bugs** — a node whose half-open range is out of bounds, empty
+   for a non-empty construct, or not nested within its parent.
+4. **Never-raise breaches** — any input that makes `lua.parse` raise instead of
+   returning `(unit, err)`.
+
+## 2. Why `load()` is the oracle (and a good one)
+
+The test harness (`tests/hull/source/test_lua_source.c`) runs co-located Lua scripts
+in a **vanilla `lua_State`** built from the **same vendored Lua 5.4** (`LUA_OBJS`)
+that Hull ships. That state's `load()` therefore drives the *exact* lexer/parser
+Hull targets — not "some Lua," the same one. Properties that make it ideal:
+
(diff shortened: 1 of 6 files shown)


farkasmark:
Four review fixes applied (commit `9cf5a2c6`) — the harness is now a gate, not just a fuzz test:

1. **Three-state Hull verdict** — `hull_state` returns `reject` / `clean` / `indeterminate`. A `lua.internal` or `lua.limit.*` marker is `indeterminate`: never read as a clean parse, never classified, and it **fails** the gate with a harness diagnostic. (Was: `hs==0` silently meant accepted.)
2. **Tightened + honest classifier** — dropped broad `too many` for precise `too many local variables`; **removed `unknown_attribute`** (Hull validates attributes → a clean parse of one is now a false-accept and fails, pinned by a direct test); `classify()` substring-matches the **normalized** message body (`chunkname:line:` stripped) and is documented as substring matching, not anchored.
3. **Complete range checks** — added `unit:text(node)` == exact slice and non-empty ranges for real constructs (only the empty-source `chunk` root may be zero-width), alongside bounds + nesting.
4. **Fail-closed enumeration** — `sl_push` latches OOM and never inserts NULL; `read_all` requires a full read; the runner **aborts** on any OOM / unreadable / short-read file (no silent skips) and reports `enumerated N, read N`.

Design record flipped to **IMPLEMENTED** and reconciled. conformance 518 over 224 files; lexer 115 + parser 133 + statements 130 + annotations 77; luacheck clean; full `hull` build OK.

PR #350, opened 2026-08-17
fix(analyze): canonical containment, traversal-pruned exclusions, honest target errors

farkasmark:
Follow-up to #349 (already merged) addressing the post-merge review: two path-safety issues + missing acceptance coverage on `hull analyze`.

## 1. Canonical containment (was lexical → symlinks could escape)
`inside_root()` compared normalized *spellings*, so an explicit `app/link.lua` symlink pointing outside the app passed containment while `path_kind()` followed it and accepted the external file. Now the root and every explicit target are **canonicalized** via a new `tool.realpath` binding before the comparison; the **logical** path is preserved for diagnostics; dedup is by canonical path. **Symlinked app roots/files are intentionally supported** (they canonicalize to their target). Also fixes the **`/` root** edge (the prefix was `//`, wrongly rejecting targets beneath root `/`).

## 2. Exclusions pruned DURING traversal (were post-filtered)
`tool.find_files` gains an `exclude_dirs` option (`cap/tool.c` threads a NULL-terminated list through `should_skip_dir`/`find_files_recurse`), so a large `build/` (or `vendor/`, `node_modules/`) tree is **never walked**. `analyze` passes `{.git,.hull,build,vendor,node_modules}`.

## 3. Honest missing vs inaccessible
`tool.realpath` returns `(nil, "missing"|"denied")` from `errno`, so a missing target is `analyze.not_found` and an inaccessible/unreadable one is `analyze.unreadable` — no longer collapsed to `not_found` by `path_kind`'s `nil`.

## Coverage
Adds `--max-depth=N` (first of the planned `--max-*` knobs) to enable a controlled **incomplete-state** test. E2E expanded to **15 cases**: symlink-inside→outside rejected, symlinked app root supported, unreadable-vs-missing (skipped as root, where perms are bypassed), non-regular target, duplicate targets, `.hull`/`node_modules` exclusions, `--json --quiet` override, incomplete state in JSON + exit 1, plus the original set.

Two new tool bindings (`tool.realpath`; `tool.stdout` from #349) + the `find_files` `exclude_dirs` thread. Full unit suite **76/76**; e2e **15/15**; conformance **520 over 225 files**; luacheck clean; full `hull` build OK.

### DIFF
--- a/docs/hull_analyze_design.md
+++ b/docs/hull_analyze_design.md
@@ -83,6 +83,17 @@ top-level `build/` and `site/build/`), plus a cheap post-filter belt. The result
 sorted, de-duplicated list of regular `.lua` files. If the root is not a readable
 directory, that is a discovery failure (exit 2), not an empty scan.
 
+**Fail closed (no silent degradation).** `tool.find_files` returns `(files, err)`; any
+allocation or traversal failure in the C walk — or a failure to build the exclusion
+list — propagates as `err` (never a partial or UNPRUNED result masquerading as
+success), and `analyze` turns it into an operational failure (exit 2). The C side
+mirrors this: `find_files_recurse` returns `-1` on any `strdup`/`realloc` failure,
+`hl_tool_find_files_ex` frees and returns `NULL` on that `-1` (and on a failed final
+grow — the old `final = results` fallback wrote one past the array), and
+`l_tool_find_files` fails closed if `exclude_dirs` was requested but could not be
+built. Likewise, **root canonicalization failure is operational (exit 2), never a
+fallback to lexical containment** (which would let a symlink escape the root).
+
 ## 4. What it does (pipeline)
 
 1. **Resolve inputs** (positional rule, §3). Walk mode → filtered `tool.find_files`.
--- a/docs/hull_analyze_design.md
+++ b/docs/hull_analyze_design.md
@@ -53,9 +53,12 @@ app root is `.` and ALL positionals are explicit files relative to `.`. So:
 
 **Flags.**
 - `--json` → machine output (one JSON object on stdout, §6).
-- `--quiet` → human-mode only: suppress the per-file "ok" chatter, print diagnostics +
-  summary. In `--json` mode `--quiet` is a **no-op** (JSON overrides it; stdout stays
-  pure JSON) — documented, not an error.
+- `--quiet` → human-mode only: suppress the summary chatter, print diagnostics. In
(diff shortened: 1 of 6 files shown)


farkasmark:
Two fail-closed fixes applied (commit `0fce1917`):
1. **Root canonicalization no longer falls back to lexical.** `tool.realpath(inp.root) or root_norm` reintroduced non-canonical containment on realpath failure (a symlink could then escape). Root-canonicalization failure is now an **operational error** (exit 2, stderr only — JSON stdout stays pure), never a fallback.
2. **Discovery propagates allocation/traversal failures** instead of a partial/unpruned "success": `find_files_recurse` returns `-1` on any `strdup`/`realloc` failure (was a silent drop) and propagates a subdir `-1`; `hl_tool_find_files_ex` frees + returns NULL on that `-1` **and** on a failed final grow (the old `final = results` fallback wrote one past the array under memory pressure); `l_tool_find_files` fails closed if `exclude_dirs` was requested but couldn't be built, and surfaces a NULL as an error (`files, err`) rather than a silent empty table (one-value callers still get the empty table — back-compat); `analyze.lua` `discover()` turns a find_files err into exit 2.

Full unit suite 76/76; e2e 15/15; luacheck clean; build OK. (OOM paths need fault injection to hit directly; verified by build + no-regression.)

farkasmark:
Traversal fail-closed fix applied (commit `f08bf596`): `find_files_recurse` now classifies filesystem traversal failures instead of swallowing them:
- `opendir`/`lstat` → **`ENOENT`** (dir/entry doesn't exist or raced away) is an explicitly-classified benign skip (also preserves `find_files`'s existing behavior on an optional/missing dir);
- **any other** `opendir`/`lstat` failure (notably **`EACCES`** — an unreadable directory, root or nested) propagates `-1`;
- `readdir` errors are distinguished from end-of-dir via `errno` and propagate `-1`.

`-1` → `hl_tool_find_files_ex` NULL → `l_tool_find_files` `(files, err)` → `analyze` exit 2. Excluded dirs are pruned *before* `opendir`, so an unreadable `.git`/`build`/`vendor` never trips this.

**E2E +1 (16 total):** an unreadable non-excluded discovered subdir → exit 2 + "discovery failed" (skipped as root). Verified no regression: `find_files` on an optional/non-existent dir still returns empty (ENOENT). 76/76 unit, 16/16 e2e, luacheck clean, build OK.

PR #354, opened 2026-08-18
feat(source): hull.source.scope — lexical binding pass (lint slice 2)

farkasmark:
Slice 2 of `hull analyze` v2 (ratified design: `docs/hull_source_scope_design.md`) — the reusable **lexical binding pass** slice 3's scope-backed rules (`unused-local`, `shadowed-local`, `unused-param`, `undefined-global`) need.

`scope.resolve(unit) -> (scope, err)`: a scope-aware recursive traversal resolving every `name` node to its binding — **local / upvalue** (across a function boundary) / **global** — with per-declaration **reads / writes / shadows**.

## Honors the Lua 5.4 subtleties
- a `local` is visible only **after** its declaration (`local x = x` RHS binds outer);
- `local function f` is visible in its own body (captured as an **upvalue** → recursion);
- params scope to the body, with implicit method **`self`**;
- loop vars scope to the body (control exprs in the enclosing scope);
- **`repeat…until`** sees the body's locals.

## Ratified tightenings
- **`function f()` is a WRITE** to the resolved `f` (local or global — not inherently global); dotted/method decls read only the **base** name.
- **same-block + same-declaration redeclaration** count as shadowing; shadowing is **same-function only** by default (upvalue shadowing not flagged); **reads-only** count as use.
- implicit `self` is **explicit** (`implicit = true` + a real anchor range).
- the chunk is **`func_id = 1`**, so nested captures of chunk locals classify as upvalues.
- the traversal is **`pcall`-guarded**: a recovered error node degrades locally, but an internal failure returns `(nil, lua.internal)` so the caller skips scope rules — never a silent partial.

Not wired into any rule (slice 3 consumes it). `test_scope.lua` (**66 cases**). lexer 115 + parser 133 + statements 130 + annotations 77 + conformance 528 + lint 67 + **scope 66**; **76/76** unit; luacheck clean; full build OK.

### DIFF
--- a/docs/hull_analyze_lint_design.md
+++ b/docs/hull_analyze_lint_design.md
@@ -183,7 +183,9 @@ is the explicit signal that lint data may be present.
    severities, `lua.lint.<id>` codes, JSON `schema_version: 2` (+ `summary.warnings`/
    `infos`/`by_rule`). `test_lint.lua` + `tests/e2e_analyze.sh` (23 cases). Lint runs
    only on a CLEANLY-parsed file (complete + no syntax errors).
-2. **Scope pass** — `hull.source.scope` + `test_scope.lua` (the reusable Step B).
+2. **Scope pass** — DONE. `hull.source.scope` (`resolve -> scope, err`; pcall-guarded)
+   + `test_scope.lua`. The reusable Step B; not yet wired into a rule (slice 3).
+   Design: [hull_source_scope_design.md](hull_source_scope_design.md).
 3. **Scope rules** — `unused-local`, `unused-param`, `shadowed-local`, and
    `undefined-global` (off by default) on top of the pass.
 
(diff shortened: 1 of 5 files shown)


farkasmark:
Two review fixes applied (commit `4c70cc90`):
1. **`ref_of` records access mode** — every resolution now carries `access = "read"|"write"` (on binding refs *and* globals), so slice 3 can distinguish a global **read** (the domain of `undefined-global`) from a global write. Tested on ordinary assignments + simple function declarations (both writes) and reads.
2. **Internal error is diagnostic-shaped** — constructed via `hull.source.diagnostic.error` with `unit.path` + an AST/unit range (severity/code/path/range), keeping the shared diagnostic contract intact (was a bare `{code, message}`).

test_scope 66 → 86; luacheck clean; full build OK.

PR #363, opened 2026-08-18
feat(frontend): JS source-frontend Slice 1 - restricted QuickJS tooling runtime

farkasmark:
Slice 1 of the JavaScript source-analysis frontend (design: `docs/javascript_source_frontend_design.md`, ratified). This slice stands up the **C-owned, restricted QuickJS tooling runtime** that will host the JS frontend, and proves it independently of any parser (Slice 2+).

Per D8 of the ratified plan, Slice 1 must prove the restricted runtime, byte transport, module loading, limits, exception conversion, lifecycle, and a trivial bundled-tool invocation. It does all of that with a throwaway `hull:probe` bundle; the real lexer/parser arrives in Slice 2.

## What lands

- **`include/hull/frontend/js_session.h` + `src/hull/frontend/js_session.c`** - a self-contained `JSRuntime`+`JSContext` that runs ONLY trusted bundled `hull:*` tooling modules from a cli-js VFS. Application JavaScript is never executed here. No `db`/`fs`/`http`/`env`/`crypto`/`compute`/`gpu`/network/spawn; `eval()` + the `Function` constructor are deleted at the JS level; bounded heap/stack/instructions/source-size/result-size. Every breach (limit, exception, bad transport, internal) becomes a structured **indeterminate** `js.limit.*` / `js.internal` diagnostic. The boundary never raises a QuickJS exception across the C seam and never crashes.
- **Byte transport** - source crosses as raw bytes via `JS_NewArrayBufferCopy` (length-aware, NUL-safe, never a NUL-terminated string); options as parsed-JSON in; frontend-neutral facts as `JSON.stringify` out. AST / `JSValue` / session handles never cross this boundary.
- **`stdlib/cli/js/hull/{probe,_probe_util}.js`** - a trivial two-module bundle registering `globalThis.__hull_frontend`, reporting the crossing (byte length, first/last byte), the security surface (present application authorities - asserts none), and an options echo. Proves multi-module loading from the tooling VFS.
- **`Makefile`** - embeds `stdlib/cli/js` as a cli-js registry (`hl_stdlib_js_cli_entries[]`), compiles the session TU, and links both into the `hull` binary gated on `RUNTIME != lua` (the tooling session needs QuickJS). **`mk/tests.mk`** builds/runs `test_js_session` (filtered out of a lua-only build).
- **`tests/hull/frontend/test_js_session.c`** - 9 cases: create/destroy (+ NULL-safe), the crossing + module loading, embedded-NUL non-truncation, empty application-authority surface, options transport, reuse across calls (entry loads once), the source-size limit, and unknown-module / unknown-method fail-closed.

## Verification

- Warning-clean under the full hull `CFLAGS` (`-Wall -Wextra -Wpedantic -Wshadow -Wformat=2` + `-Werror=...`).
- `test_js_session` 9/9 green.
- Green under **ASan + UBSan** (clean lifecycle: create/destroy, 50x reuse).
- Full `make` (hull binary) links with the gated session + cli-js registry.

## Scope / non-goals

No parser, no annotations, no scope, no Query/Compute IR, no C dispatcher binding, no application-JS execution. Those are Slices 2-7, each a separate PR per the ratified plan.


farkasmark:
Pushed `2bbdd583` addressing all six blockers.

**1. Null source with nonzero length (critical).** `hl_js_session_analyze` now rejects `src == NULL && src_len > 0` before entering QuickJS, returning `js.transport`. Covered by `null_source_nonzero_len_rejected` (and `null_source_zero_len_ok` proves a genuine empty source still works); run under ASan in CI.

**2. Malformed options fail closed + length-aware.** The API is now `(options_json, options_len)` and options are `JSON.parse`'d with the explicit length. Malformed / embedded-NUL / trailing-garbage input returns a structured `js.transport` diagnostic (was: silently nulled; `strlen`-based parsing also truncated at an embedded NUL). Pending parse exception is cleared correctly. Covered by `options_malformed_fails_closed`, `options_embedded_nul_fails_closed`, `options_trailing_garbage_fails_closed`.

**3. Dynamic code actually blocked.** This is the substantive change. Both C `JS_Eval` and the `Function` constructor funnel through `ctx->eval_internal` (set only by `JS_AddIntrinsicEval`), and QuickJS makes eval optional - with the hook NULL, *every* dynamic-compile path throws "eval is not supported", including the prototype-reachable constructors. So the session context **never** enables the eval hook. The trusted cli-js bundle is compiled to bytecode once in a throwaway eval-enabled context at create, and the session loads it via `JS_ReadObject`/`JS_ResolveModule`/`JS_EvalFunction` (no eval hook needed). Global `eval`/`Function` are still deleted as defense in depth. `dynamic_code_blocked` adversarially asserts `global_eval`, `global_Function`, `({}).constructor.constructor(...)`, `(function(){}).constructor(...)`, and `(async function(){}).constructor(...)` are all blocked and nothing `RAN`. No JS poisoning and no vendored-QuickJS patch.

**4. Full limit contract exercised + reclassified.** New tests: `limit_source_size` (bytes), `limit_instructions`, `limit_stack`, `limit_heap`, `limit_result_size`, and `ordinary_throw_is_internal`. Heap/stack breaches are recognized from the QuickJS message and classified `js.limit.heap` / `js.limit.stack` (were falling into `js.internal`). The header now documents that `max_instructions` counts interrupt-handler invocations - one per `JS_INTERRUPT_COUNTER_INIT` (10000) interpreter poll-points, not literal instructions - and that the interrupt error is uncatchable.

**5. Null-output leak fixed.** `fail_indeterminate` now frees the built diagnostic when `out_json == NULL`, matching the success path. `null_output_lifecycle` covers both the failure and success no-output paths.

**6. Argument-construction failures checked.** `JS_NewArrayBufferCopy` / `JS_NewStringLen` are checked for exceptions before `JS_Call`; a construction failure surfaces as `js.transport` with full cleanup instead of passing exception values onward.

The session no longer uses `HlVfs` (it iterates the cli-js registry directly and loads by bytecode), so the test link dropped `vfs` / `path_normalize` / `sh_seal_arena`. 21 cases; warning-clean; green under ASan + UBSan; full hull binary links.

farkasmark:
Pushed `e4dbb9ac` addressing the four remaining boundary issues.

**1. Options pointer/length mismatch.** `hl_js_session_analyze` now rejects `options_json == NULL && options_len > 0` as `js.transport`, symmetric with the source check, before entering QuickJS. Covered by `options_null_ptr_with_len_rejected`.

**2. Single per-invocation work budget + phase-uniform classification.** The instruction counter is reset **once** at invocation entry (was: reset immediately before `JS_Call`, leaving options parsing, argument construction, the one-time entry load, module jobs, and serialization under a stale or unintended budget). A new `fail_from_exception` helper consumes the pending exception and classifies it uniformly across `JS_ParseJSON`, argument construction, module-job draining, `JS_Call`, `JS_JSONStringify`, and `JS_ToCStringLen`, so a resource breach reports the same `js.limit.*` code regardless of which phase tripped it.

**3. Argument-construction OOM is `js.limit.heap`, not transport.** A valid source that cannot fit the configured QuickJS heap during `JS_NewArrayBufferCopy` now classifies as `js.limit.heap` (the pending exception is preserved and inspected before clearing), not malformed transport. Tested deterministically by `limit_heap_large_source` (16 MiB source, 8 MiB heap), which asserts `js.limit.heap` and NOT `js.transport`.

**4. The result is required to be JSON.** `JSON.stringify(undefined)` yields the JS value `undefined` (not a string); `JS_ToCStringLen` would otherwise emit the bytes `undefined` and the boundary would return it as success. `analyze` now requires `JS_JSONStringify` to return a string and fails closed (`js.internal`) otherwise, so Slice 1 never knowingly emits non-JSON. New probe method `returnsUndefined` + test `result_undefined_is_indeterminate`. (A complete C-side JSON parse remains a dispatcher-slice concern, as you noted.)

`classify_exception` is replaced by `fail_from_exception`. 24 cases; warning-clean; green under ASan + UBSan; full hull binary links.

farkasmark:
Pushed `5f99e1c2` addressing the three residual exception-path issues. All are on the failure edges, so no invocation can leave a pending exception that contaminates the next call on a reused session.

**1. Secondary pending exception in `take_exception_message`.** When `JS_ToCString` itself fails (most relevant under OOM), it returns NULL *and* raises a new pending exception. That secondary exception was never cleared. It is now cleared before returning the fallback message. New `reuse_after_heap_failure` test: trip a heap breach (`js.limit.heap`), then a normal `analyze` on the same session must succeed with `status:ok`, proving no stale exception carried over.

**2. `drained != 0` ordering in `ensure_entry_loaded`.** The rejected-promise branch ran before the drain-failure check, so a pending-job failure from an interrupt / heap / stack breach could be labeled `js.internal`. `drained != 0` is now handled first through `fail_from_exception` (unified `js.limit.*` classification); the ordinary promise-rejection branch is only reached after a successful drain.

**3. Frontend / method lookup outside unified handling.** A `JS_GetPropertyStr` that throws (e.g. OOM) was interpreted as "entry did not register" / "method is not a function", leaving the OOM exception pending. Both reads now check `JS_IsException` and route through `fail_from_exception` first.

25 cases; warning-clean; green under ASan + UBSan; full hull binary links.

farkasmark:
Pushed `afe7ed8b` addressing the final exception-path edge.

The rejected-promise branch in `ensure_entry_loaded` called `JS_ToCString(res)` directly, so a conversion failure under memory pressure raised a secondary pending exception that was never cleared, recreating the contamination just fixed for `take_exception_message`.

Extracted `jsvalue_to_message(ctx, val, fallback)`: it converts any `JSValue` to a best-effort heap string and clears the secondary exception `JS_ToCString` raises on failure. Both `take_exception_message` (pending exceptions) and the rejected-promise branch (rejection values) now go through it, so neither can leave a stray pending exception on a reused session.

New `hull:_probe_reject` tooling module (rejects on load via top-level await) + `entry_rejection_and_reuse` test: the entry-rejection path reports `js.internal` and the session still serves a subsequent good-entry call, exercising the previously-untested rejected-promise branch and proving reuse stays clean.

26 cases; warning-clean; green under ASan + UBSan; full hull binary links.

farkasmark:
CI status on current head `afe7ed8b`: **48 pass, 1 skipped, 5 fail**. All Slice-1 code is green (build across all platforms, ASan+UBSan, and `test_js_session` runs clean under MSan too). The 5 reds are two pre-existing, main-reproducible failures unrelated to this PR:

**1. Project source-discovery E2E (`hull agent inspect`)** - fails on Linux, Linux (aarch64), macOS, Linux (clang). Also fails on `main` (run 32147117016, commit 904488ce, same 4 platforms). Passes locally here (51/51). It exercises the Lua tool-VM `hull dev --agent` path; zero overlap with the JS tooling runtime added in this PR.

**2. MSan + UBSan timeout** - the `MSan test` step times out after 15 minutes deep in the existing `test_js` suite's TOTP tests (`totp_enroll_confirm_verify` ~47s, `totp_pending_cleanup` ~34s under MSan instrumentation). `main` times out at the *exact same* test (`js_stdlib.totp_brute_force_lockout`, run 32147117016) without `test_js_session` even existing, so it is a structural pre-existing budget overrun, not introduced here. `test_js_session` itself runs fully green under MSan (all 26 cases `[ OK ]`) before the unrelated timeout.

Neither failure is attributable to Slice 1. Per the review guidance, holding the merge and treating both as separate CI blockers rather than folding an unrelated fix into this PR.

PR #445, opened 2026-08-31
chore(keel): migrate to Keel v3.0.0-rc.3

farkasmark:
Refreshes the vendored Keel HTTP server library from **v2.7.1 → v3.0.0-rc.3** and adapts Hull to the v3 public HTTP surface. This is **Slice 1** of the Keel-v3 work: the bump as a standalone, zero-client-behavior-change PR. The client-unification slices (SMTP spike → Postgres/MySQL wire onto Keel transport) are separate follow-on PRs.

Rename-only across the tree except **three** real changes (opaque-conn peer addr, the async send-completion re-integration, and the vendored `sh_seal_arena`).

## What changed

- **HTTP taxonomy re-prefix.** Headers → flat `keel/http_*`; types → `KlHttp*`/`KlHttp1*`/`KlHttp2*`; functions → `kl_http_*`; macros → `KL_HTTP_*`. `ws`/`url`/`error`/`async`/`timer`/`allocator`/`tls` headers are unchanged in v3.
- **Backend adapters moved to `integrations/`.** mbedTLS → `integrations/tls/mbedtls/`, miniz → `integrations/codec/miniz/`. Still built via `KEEL_TLS=mbedtls` / `KEEL_COMPRESS=miniz`; added the integration `-I` paths + `MBEDTLS_CONFIG_FILE`.
- **Opaque `KlHttpConn`.** `req.remote_addr` now goes through `kl_http_conn_peer_addr` + `kl_sockaddr_format_ip` (drops `getpeername(conn->fd)`); request fields read directly off the concrete `KlHttpRequest` struct.
- **Async send-completion re-integrated to the v3 contract.** Hull suspends, builds the response, and `kl_async_complete` lets **Keel own the send** (incl. partial/would-block) — Hull no longer drives the connection state machine. Deletes the manual `conn->state` / write-loop code.
- **Two new v3 public entry points, surfaced by Hull dogfooding.** The streaming-async multipart path resumes from the body reader's `on_data` (not `kl_async_complete`), which had no public v3 mechanism after the F2 conn-opacity refactor. Keel added:
  - `kl_http_request_await_body(req)` (rc.2) — "park, keep reading the body"
  - `kl_http_request_send_response(req)` (rc.3) — "handler done, transition to sending"
- **Vendored `sh_seal_arena` into Hull** (`vendor/sh_seal_arena/`). v3 no longer ships it, but Hull's `hl_seal_arena` (the §4b manifest/VFS sealer) depends on it.

## Security note (docs/security.md §4e)

Request sealing (`KEEL_SEAL_REQUEST`) and router freeze (`kl_server_freeze`) — two mprotect-based defence-in-depth layers — were **removed by Keel v2.8.0** (not v3). Hull's pre-2.8.0 pin crossed the removal. New §4e documents the loss and the residual exposure. **Primary enforcement is unchanged**: the `hl_cap_*` capability boundary, the kernel sandbox, the §4b sealed manifest, and §4c binary hardening all stand. A Hull-side re-implementation via `hl_seal_arena` is recorded as a future enhancement.

## Validation (on rc.3)

| Check | Result |
|---|---|
| Default `make` | ✅ |
| Unit suites | ✅ 91/91 |
| e2e-http (async / SSE / ws) | ✅ |
| e2e-multipart | ✅ 94/94 |
| e2e-examples (14 apps × Lua+JS) | ✅ 206 / 0 |
| Flavor: pure-compute (`HL_ENABLE_HTTP=0`) | ✅ link |
| Flavor: sqlite+postgres | ✅ link |
| Flavor: sqlite+mysql | ✅ link |
| No surviving Keel 2.x names | ✅ |
| Cosmo (`CC=cosmocc`) | ⏳ **CI-validated** — cosmocc not installed on the dev box |

## Follow-on (separate PRs, not this one)

1. **SMTP spike** — prove Keel v3's `KlStream`/`KlSocketProvider` + async DNS can back a synchronous blocking client; migrate `cap/smtp.c` off raw sockets.
2. **Postgres / MySQL wire** onto Keel transport (conditional on the spike result).


farkasmark:
Pushed `cf9bc34c` addressing all four blockers. Two of the CI failures were **real bugs my local macOS build masked** — thanks for catching them.

### 1. CI matrix (red → root causes fixed)
Two genuine bugs, plus the fast gates:

- **`sh_seal_arena` undefined when composing apps** (the Linux/macOS/feature/reproducible/HTMX/spans failures). `vfs.o` is in `libhull_platform.a` and calls `sh_seal_arena_*`; Keel v3's `libkeel.a` no longer bundles that symbol, and the Makefile only added Hull's own `sh_seal_arena.o` to the archive in the **no-keel (pure-compute)** case — full builds relied on the keel-merge. Fix: `sh_seal_arena.o` is now added to the platform archive **unconditionally**. The `hull` *binary* linked fine (its own link line already lists it), which is why my local builds passed — only the composed-app archive was short. Validated: `ar t libhull_platform.a | grep sh_seal_arena` = 1 for both full and pure-compute, and `e2e-feature-runtime` passes.

- **Cosmo timeout was a real regression, not flakiness.** It hung on the JS `webhooks` example's self-fetch (`POST /events` → `outbox.flush()` → `http.fetch` to `127.0.0.1`). Root cause: an asymmetry my migration introduced — Lua's async resume calls `kl_http_request_send_response(req)` (the v3 successor to the old `conn->state = KL_CONN_SENDING`) in the FULFILLED + error paths, but JS's `hl_js_async_resume` did not. Under kqueue/epoll `kl_async_complete`'s immediate send happens to flush; under **poll** (cosmo) it does not drive a resumed handler's send, so the response never went out. I reproduced it **natively** by building Keel with `BACKEND=poll` and relinking (JS hung, Lua was fine), added the call + `js->active_req` restore to match Lua, and confirmed the fix under poll (`POST /events` 200 in ~1 ms). The Cosmopolitan job was green on `main` including this PR's merge-base, so it was my regression.

- **Lint / sign-off**: the 4 em-dashes are gone (`check-no-emdash` passes), and the commit is now `Signed-off-by`.

### 2. `install.sh.minisig`
This PR does not touch it — it's byte-identical (`74a741b46359`) across the merge-base, `origin/main`, and HEAD, and it's absent from `gh pr view 445 --json files`. There's no signature churn in the diff to remove; whatever regenerated it was outside this branch.

### 3. Seal-arena provenance / stale claims
Corrected. `vendor/sh_seal_arena/{h,c}` now record the provenance explicitly (vendored from **Keel v2.7.1**, MIT, © Artalis, `SPDX-License-Identifier: MIT`) and state Hull is the **sole** consumer; removed the stale "Keel uses it via `kl_router_freeze` / per-request snapshot" claims. Fixed the two Makefile comments that named the old `vendor/keel/vendor/sh_seal_arena/` path (now `vendor/sh_seal_arena/`).

### 4. Security wording (§4e)
Tightened. Dropped the overstated "request method/path/headers are re-validated at each capability call" claim. The section now draws the narrower distinction you asked for: the sealed-manifest **capability grants** (`fs`/`hosts`/`env`) still can't be widened by a corrupted request, **but** application authentication, authorization, routing, and middleware read request data, and the removed in-place-rewrite seal was a real (narrow) defence-in-depth reduction against a memory-corruption primitive.

### Local validation on `cf9bc34c`
`make test` 91/91 · `test_js` 157 · `test_lua` 172 · `e2e-multipart` 94/94 · `e2e-examples` 280/0 (JS webhooks 9/9) · `e2e-feature-runtime` compose · pure-compute + full archives both self-contain `sh_seal_arena` · JS poll-backend self-fetch reproduced and fixed. Re-running the full matrix now.

farkasmark:
### Exact-head CI is green at `24991dd0`

Full matrix: **64/64 jobs pass, 0 fail** (run 33441345227). The gates you named:

| Gate | Result |
|---|---|
| CI success (applicability-aware gate) | ✅ |
| Cosmopolitan (APE) | ✅ |
| MSan + UBSan | ✅ |
| Sign-off check (DCO) | ✅ |
| Lint (em-dash) | ✅ |
| All 7 build flavors (pure-compute, compute-only, image-less, mysql-only, postgres-only, sqlite+mysql, sqlite+postgres) | ✅ |
| Feature builds (DuckDB, GPU, MySQL, Postgres, TUI) | ✅ |
| Linux / Linux (aarch64) / Linux (clang) / macOS | ✅ |
| Reproducible (Linux, Linux container, macOS, Cosmo a/b/compare) | ✅ |
| HTMX browser (Linux + macOS) | ✅ |

**One more real regression surfaced and was fixed this round** (the macOS CA-bundle failure, which I reproduced locally — not a network flake):

- **Async HTTPS returned nil / `Bus error`.** Sync HTTPS worked (TLS + mbedtls integration fine), and async-to-`127.0.0.1` worked, but async-to-a-hostname failed with `"DNS resolution failed"` (visible only under `--audit`). Keel v3 flipped the async client's DNS default: `KlHttpClientConfig.system_dns=0` now selects a **built-in async DNS resolver** that reads `resolv.conf` / sends UDP directly — which Hull's kernel sandbox blocks (it grants `network_outbound`, so `getaddrinfo` via the OS resolver is fine, but not the direct path). Fix: `system_dns=1` on the async path (`cap/http_async.c`), restoring the pre-v3 blocking-`getaddrinfo` resolution. The sync client is always blocking, so it was unaffected.

The other round-2 failure (`Reproducible build (Linux container)`) was a transient Ubuntu apt-mirror 500 during `docker build` — it's green on this run.

The diff is ready for your focused re-review.

PR #447, opened 2026-09-01
spike(smtp): Slice 2a feasibility spike (review-only, DO NOT MERGE)

farkasmark:
**Review-only feasibility spike — not for merge.** Slice 2a of the SMTP→Keel-v3 transition ([design PR #446](https://github.com/artalis-io/hull/pull/446)). Throwaway Hull-local code under `spike/`; **no production source touched** (`cap/smtp.c` unchanged), **no shared transport abstraction**. Stops here for API + ownership review per the design.

## Question it answers
Can an SMTP-shaped transport adapter compose Keel v3's public primitives (connect → byte stream → TLS upgrade → close) using **only public Keel headers** — no HTTP-private/`src/` header, no combined facade? And if not, what narrow primitive is missing?

## Verdict: YES (on POSIX), no blocking missing primitive

- **Public-header compile gate holds.** Compiles `-Wall -Wextra` clean with only `-Ivendor/keel/include` + the one TLS integration dir; **no `-Ivendor/keel/src`**. `nm` confirms zero private `kl_sock_*` consumer wrappers and no HTTP symbols.
- **Lifecycle runs green** over one `KlEventCtx` (macOS kqueue readiness): connect via `KlConnectOp` → `220` greeting → `EHLO` → coalesced `250` (asserted on the *accumulated* reply, proving fragmented/coalesced read handling) → `QUIT` → `221` → graceful close. Connect **cancellation → clean detachment**. `on_done`/`on_detach`/`on_close` each fire **exactly once**. **ASan clean** (no leak/UAF; fd closed after detach).
- **Socket I/O** reached via `kl_socket_provider_posix()->ops->*` (fully-populated public vtable) directly — the private `kl_sock_*` wrappers are not needed.
- **TLS**: wiring compile-verified against the public `KlTls` vtable + `keel_tls_mbedtls.h` (STARTTLS handshake-loop structure); live handshake deferred (needs a TLS peer).

## Narrow seams noted (provider-neutrality only, not POSIX gaps)
1. POSIX provider leaves `ops->io_status` NULL (errno fallback) — a freestanding/non-POSIX (lwIP/UEFI) adapter would need providers to supply it.
2. `KlEventCtx` exposes no public socket-provider accessor — matters only if a future adapter must inherit a non-POSIX provider from the ctx.

[...]

PR #449, opened 2026-09-01
feat(smtp): Slice 2b - route SMTP transport through Keel v3 primitives

farkasmark:
Slice 2b of the SMTP-to-Keel-v3 transition (design: `docs/smtp_keel_client_design.md` §5/§6/§7/§9; Slice 2a spike = closed PR #447). **Transport-only rewrite** of `cap/smtp.c`; protocol, policy, tokens, audit, pure helpers, and the wire conversation are preserved. **Stops here for protocol/security review** (Slice 2b boundary); Slice 2c (worker/runtime scheduling) is not started.

## What changed
- **New `cap/smtp_transport.c` + header** — a Hull-local adapter composing Keel v3 **public** primitives (`KlConnectOp` resolve+Happy-Eyeballs connect, `KlStream` bounded writes / strict read pause-resume / graceful close, `KlTls` implicit + in-place STARTTLS, timers, POSIX socket-provider public ops) on **one private, operation-local `KlEventCtx`** the synchronous `hl_cap_smtp_send` pumps to terminal (design §6 model 1). Public headers only (no `vendor/keel/src`, no private `kl_sock_*`, no facade; `*_detail.h` only to embed as fields).
- **`cap/smtp.c`** — deleted `smtp_connect`/`io_read`/`io_write`/`smtp_read_line`/`smtp_read_response`/`smtp_send_command`/`smtp_send_raw` and all raw-socket includes; the state machine now drives the adapter. Added an **incremental reply parser** (partial-byte retention, multi-line-per-read, consistent continuation code, bounded).
- **Makefile** — `smtp_transport.c` rides `CAP_OBJS`; added to the `HL_ENABLE_HTTP_CLIENT=0` filter so a client-less base drops it with `smtp.c`.

## Preserved exactly
`hl_cap_smtp_send` signature + 0/-1 return; the Lua/JS `smtp.send` C ABI (`mod_smtp.c` untouched); the pure helpers; **every** `err_msg` token at its original decision point; the audit record; the EHLO/MAIL/RCPT/DATA/QUIT conversation + codes.

## Security invariants (verified in review)
Host auth **before** resolve/connect (against the declared hostname; connected address never widens it) · CR/LF rejected · **AUTH refused unless TLS active + verified** · **STARTTLS no-downgrade** (rejection or handshake failure is terminal, never plaintext) · TLS CA+hostname verify via `cfg->tls` · audit = metadata + result only. Plus the two Slice-2a invariants: **persistent allocator** outlives every `KlTls`/`KlTlsCtx`; **STARTTLS pauses plaintext reads + hands the fd to a dedicated handshake watcher before any plaintext read can consume ClientHello bytes**.

## Acceptance
`test_smtp` 38/38 · `test_smtp_e2e` 8/8 · `test_http` 9/9 (shared `tls_client` unaffected) · ASan-clean across connect/TLS/close. (Full CI on push.)

## Flagged for the protocol/security review
1. **Credential scrubbing not added** — the AUTH `plain`/`b64` stack buffers are not `secure_zero`'d after use. This is **parity** with the pre-2b code (it never scrubbed either); the design lists scrubbing under the *final* transition (§10). Recommend a dedicated hardening (here or a follow-up) rather than silently expanding this transport swap.
2. **Resolution is a blocking `getaddrinfo`** inside the adapter's `start_resolve` (never in `cap/smtp.c`), completing inline and handing the address list to `KlConnectOp` to race IPv4+IPv6. Deliberately preserves the sandbox-compatible system resolver (`/etc/hosts`, search domains, the network-outbound grant) - the same reason `http_async` forces `system_dns=1`. Net gain: multi-address racing vs the old first-address-only connect.
3. **Deadline model** — one per-stage budget via the pump loop + the `KlConnectOp` connect-deadline timer, mapping a timeout to the caller's existing per-stage token. Satisfies the tests; the design's "total + independently-bounded stage deadlines" (open question 1) is a straightforward extension for a later pass.
4. **Minor**: on an OOM partial-`kl_stream_*` bring-up failure the write queue could leak (error path only, pre-`stream_up`); noted, not a happy-path concern.


farkasmark:
### Review blockers addressed (`cca9bf73`)

All 11 fixed under `cap/smtp.c` + `cap/smtp_transport.{c,h}` + a new focused test. Verified by me (code review of each fix + tests + ASan), not just the builder's report.

| # | Fix | Verified |
|---|-----|----------|
| 1 | Message-size: **bounded chunked admission** (256 KiB) with ordered drain; 10 MiB message no longer hits `KL_STREAM_TOO_LARGE` | `hl_smtp_transport_write` rewritten; test `chunked_delivery_10mib` |
| 2 | Connect ownership: teardown `kl_connect_op_cancel` + pump to `is_detached` before freeing; **fails loud (no free)** if it won't detach | `hl_smtp_transport_free`; test `blackhole_timeout_detaches` |
| 3 | Event-loop args were reversed → `kl_event_ctx_run(ev, 64, step_ms)`; **monotonic absolute deadline** | `pump_until` |
| 4 | **Real Happy Eyeballs**: `arm_delay`/`cancel_delay` schedule a 250 ms RFC 8305 stagger → `kl_connect_op_on_delay` | `co_arm_delay`/`co_cancel_delay` |
| 5 | Reply parser: **exact terminators only** (`NNN-` / `NNN␠` / bare `NNN␍␊`; any other 4th char rejected); total bound enforced **before every line incl. terminal** | `reply_acc_take`; 9 parser tests |
| 6 | STARTTLS: **require empty accumulator after the 220** (unexpected bytes abort fail-closed); **no `kl_stream_resume` on any TLS failure** | `hl_smtp_transport_starttls` + `tp_tls_abort`; test `unexpected_buffered_bytes_aborts` |
| 7 | TLS completeness: validate vtable, check `set_hostname` + `kl_stream_resume` returns, don't report success if reattach fails, remove handshake watcher before destroy/close on timeout/cancel | `tp_tls_begin_handshake`/`_drive`/`_abort` |
| 8 | Partial stream-init leak: `write_q_inited` frees the queue whenever initialized | `hl_smtp_transport_free` |
| 9 | **Credentials scrubbed** (this PR): single-cleanup AUTH helper secure-zeros raw/base64/command on every path; creds stay borrowed | `smtp_do_auth_plain`; test `secure_zero_clears_buffer` |
| 10 | **CRLF validation now covers `content_type`** | `smtp_validate_message` |

Dispositions applied: blocking `getaddrinfo` stays for 2b parity **with a comment that the per-stage deadline does not bound DNS and Slice 2c must move it off-thread**; the hard total-operation ceiling is a 2c item.

**Acceptance:** `test_smtp` 38/38 · `test_smtp_e2e` 8/8 · `test_smtp_transport` **15/15** · ASan+UBSan clean.

### Honest test-coverage caveats (flagged for you; I recommend deferring to Slice 2c)
1. **Happy Eyeballs stagger has no behavioral test** — a deterministic *delayed-winning-address* test needs a controllable dual-A/AAAA resolver, which the (unchanged, sandbox-parity) blocking `getaddrinfo` doesn't expose. The hooks are wired + compile-verified; the Slice-2a spike demonstrated multi-address racing live. Natural to land the behavioral test in 2c behind the mockable off-thread resolver.
2. **Blackhole test proves clean detachment on connect-failure, not necessarily the deadline *timer* firing** — `192.0.2.1` returns immediate `EHOSTUNREACH` on this host rather than hanging to the deadline. A true deadline-fire test needs a DROP endpoint CI can't guarantee.
3. **TLS failure branches (vtable/hostname/reattach/timeout) are code-review-verified, not live-integration-tested** — driving them needs a full in-process mbedTLS peer (the spike pattern); the natural home is 2c alongside real STARTTLS/implicit-TLS handshake coverage.

Stopping for your protocol/security **re-review** + the full cross-platform matrix (CI on push). Slice 2c remains stopped.

farkasmark:
### Review round 2 addressed (`5f79a387`)

All four remaining blockers + the teardown-observability concern fixed and **fix-proven** (each fix was temporarily reverted to confirm its new test fails/crashes, then restored):

| # | Fix | Proof |
|---|-----|-------|
| 1 | **Invalid TLS vtable no longer crashes on rejection** — the reject path NULL-checks `destroy` before calling it (a NULL `destroy` is *why* `kl_tls_vtable_valid` fails) | new death test `malformed_vtable_rejected_no_crash` (segfaults if the guard is removed) |
| 2 | **Missing `set_hostname` no longer silently accepted** — a non-empty host now **requires** the hook to exist *and* succeed, else abort (SNI/verification can't be skipped) | `missing_set_hostname_rejected` + `failing_set_hostname_rejected` |
| 3 | **One write stage = one absolute deadline** — `pump_until_abs`; `hl_smtp_transport_write` computes the deadline once and shares it across every chunk drain + WOULD_BLOCK retry (a slow peer can't stretch a 10 MiB write to N×timeout). Total-operation ceiling stays a 2c item | code + existing chunk tests |
| 4a | **STARTTLS buffer test proves the right thing** — now uses a **valid** mock TLS config and asserts the **factory was never called**, so the -1 is attributable to the non-empty accumulator, not a NULL cfg | reworked `unexpected_buffered_bytes_aborts` |
| 4b | **Credential test drives the real helper** — a gated test seam (`HL_SMTP_TEST_HOOKS`, **absent + zero-overhead in production**) lets the test exercise `smtp_do_auth_plain`'s real success **and** failure exits; the send hook proves the live AUTH PLAIN was built, a post-cleanup probe proves `plain`/`base64`/`command` are all zeroed on both paths | `auth_plain_success_scrubs` + `auth_plain_failure_scrubs` |

**Teardown observability:** `hl_smtp_transport_free` now returns `int` (0 = freed; -1 = a `KlConnectOp` won't detach, so storage is intentionally leaked rather than freed into a use-after-free). `hl_cap_smtp_send` observes it, logs at the capability boundary, and records `"teardown":"leaked"` in the audit — no longer silently dropped.

**Verification:** `test_smtp` 38 · `test_smtp_e2e` 8 · `test_smtp_transport` **19** · ASan+UBSan clean · seam compiled out of the shipped binary (`nm cap_smtp.o` → no `smtp_test_auth`). Production path unchanged (the `#ifdef` only adds the hook; the scrub runs on all paths regardless).

Stopping for your re-review + the full cross-platform matrix. Slice 2c remains stopped.

farkasmark:
### Review round 3 addressed (`93dae25a`)

Both remaining blockers + the observability gap fixed:

**1. Stream detachment enforced before free.** `hl_smtp_transport_free` applied the fail-closed rule to the connect op but ignored the stream cancel/pump result. It now verifies `t->closed` after the abortive `kl_stream_cancel` + pump; if the stream doesn't reach confirmed detachment within the bound it **returns -1 and intentionally leaks storage** rather than destroying TLS / closing the fd / freeing the event ctx + `t` while live stream callbacks may still reference them — the same rule as the connect op.

**2. Shared absolute write deadline now has a behavioral regression test.** New `smtp_write.write_deadline_is_one_absolute_bound`: a peer greets then **stalls forever**; a 10 MiB (~40-chunk) write with a 500 ms timeout returns -1 in **well under 2 s** (asserted), where the pre-fix per-chunk budget would have needed up to ~20 s. (Reverting to per-chunk `pump_until` makes it hang and blow the 2 s bound.)

> Writing this test surfaced a real teardown footgun I also fixed: a **graceful** `hl_smtp_transport_shutdown` of an undrainable queue (dead/stalled peer) waited the full 30 s operation timeout. Now bounded to a short grace (`SMTP_CLOSE_GRACE_MS`), with the abortive free finishing teardown — and `hl_cap_smtp_send` skips the graceful close entirely on its failure path (a failed send goes straight to the abortive free).

**Observability gap.** `hl_smtp_transport_connect` gained an `out_teardown_leaked` parameter (the connect path is the most plausible non-detaching-op site). `hl_cap_smtp_send` declares `teardown_leaked` before the connect, routes the connect-failure through the **single** cleanup+audit block, and ORs the connect outcome with the cleanup-free result — so a connect-path teardown leak also reaches the `"teardown":"leaked"` audit field.

**Verification:** `test_smtp` 38 · `test_smtp_e2e` 8 · `test_smtp_transport` **20** (whole suite ~4 s) · ASan+UBSan clean · production build clean with the AUTH test seam still compiled out (`nm cap_smtp.o` → no `smtp_test_auth`).

Stopping for your re-review + the full cross-platform matrix. Slice 2c remains stopped.

farkasmark:
## Round 4 — the write-deadline regression now genuinely bites

Head: \`2f92326e\` (fast-forward on round-3 \`93dae25a\`).

**The gap you found was real.** The round-3 test could pass against the reverted code:
- The stall peer closed at 1.5 s; that close sets \`read_eof\`, which terminates *both* the fixed absolute-deadline write and the pre-fix per-pump-budget write under the 2 s ceiling.
- I also confirmed a *fully-stalled, never-draining* peer (your suggested "open and unread until released") does **not** distinguish them either: the true pre-fix code (\`cca9bf73\`) returns \`-1\` as soon as its **first** fresh \`timeout_ms\` budget expires — it does not hang. So a black-hole peer + watchdog would have passed on the bug too.

**What actually separates fixed from reverted: a slow-but-steady drainer.** Each ~256 KiB chunk drains inside one fresh budget, so the pre-fix code keeps *resetting* the deadline and **delivers the whole 10 MiB (returns 0)** after ~5 s; the fixed single absolute deadline trips mid-write and **returns -1 at ~one budget**.

New test design:
- Peer reworked to a **deterministically throttled drainer**: sleeps a fixed time per fixed byte quantum (64 KiB / 30 ms → ~2 MB/s), independent of read granularity, so the rate holds regardless of loopback speed. It reads until EOF, so the **client** decides when it goes away (by closing its fd), not a timer.
- Asserts **both** facets: \`rc == -1\` **and** \`elapsed < 2000 ms\`.
- **SIGALRM watchdog (30 s)** kept as a finite backstop so a *different*, genuinely-hanging future regression fails CI rather than hanging forever.
- Failure path now calls \`hl_smtp_transport_free()\` **directly** (no \`hl_smtp_transport_shutdown()\`): a graceful drain of an undrainable queue is not what a failed write does, so the test exercises the production failure teardown. \`free()\` closing the fd gives the peer EOF and the thread joins.

**Exact-head revert proof (on \`2f92326e\`):**
- Restore the pre-fix per-pump budget (\`pump_until(t, done_write, timeout_ms)\` at the 3 write-loop sites, matching \`cca9bf73\`) → test **FAILS**:
  \`\`\`
  tests/hull/cap/test_smtp_transport.c:723: Failure
    Expected : (rc) == (-1)
      Actual : 0 vs -1
  [  FAILED  ] smtp_write.write_deadline_is_one_absolute_bound (4804073166ns)
  \`\`\`
- Fixed code → **PASSES** at ~0.5 s. All 20 transport cases green.

**Also in this commit (branch-green, comment-only):** recast the \`Slice 2b\`/\`Slice 2c\` milestone-narration in the SMTP code comments (\`smtp.c\`, \`smtp_transport.c/.h\`, the test header) to present tense so \`check-no-milestone-narration\` passes. No transport behavior changes; the design doc keeps its Slice framing (\`*.md\` is gate-exempt).

**Verification:** \`test_smtp\` 38 / \`test_smtp_e2e\` 8 / \`test_smtp_transport\` 20 green; ASan+UBSan clean; normal build clean with the test seam absent (\`nm cap_smtp.o | grep smtp_test_auth\` → empty); \`check-no-emdash\` + \`check-no-milestone-narration\` + \`git diff --check\` all pass.

Awaiting the exact-head cross-platform matrix.

farkasmark:
## Exact-head matrix on 6e95c91c: one real failure (MSan), now fixed at 860ae26f

The 6e95c91c matrix came back with **every required gate green except MSan + UBSan** (which the applicability-aware gate aggregates): Cosmopolitan (APE) ✅, both reproducible Cosmo builds ✅, Static Analysis/cppcheck ✅, Lint ✅, ASan+UBSan ✅, all flavors ✅.

**The MSan failure was real and traced to my code**, not a flake. In `test_smtp_e2e` and `test_smtp_transport`:
\`\`\`
use-of-uninitialized-value
  #0 psx_get_so_error socket_posix.c
  #1 sp_get_so_error   src/hull/cap/smtp_transport.c:93
  #2 co_connect_watcher src/hull/cap/smtp_transport.c:735
SUMMARY: MemorySanitizer: ... socket_posix.c in psx_get_so_error
\`\`\`

**Root cause: the uninstrumented-library boundary, not a defect.** The `msan` target builds `libkeel.a` with Keel's own Makefile *without* MSan (the same deliberate stance Hull takes for WAMR); Hull instruments every read-path TU it compiles itself. Keel's `kl_sockdef_get_so_error` is correct (`int err=0; socklen_t len=sizeof(err); getsockopt(fd,SOL_SOCKET,SO_ERROR,&err,&len)`), but because the store to `len` happens in uninstrumented code, MSan's `getsockopt` interceptor sees the length operand as poisoned. The pre-Slice-2b raw-socket SMTP did this getsockopt in *instrumented* Hull code, so it never crossed the boundary — my transport is the first `make test` suite to drive a real non-blocking connect through Keel's socket provider under MSan (**main's MSan run shows zero `psx_get_so_error` hits and passes**).

**Fix (test-infra only, no production/design/Keel change):** `tests/msan_suppressions.txt` scoped to that one interceptor site (`interceptor_via_fun:psx_get_so_error` / `kl_sockdef_get_so_error`), wired via `MSAN_OPTIONS=suppressions=…` in the `msan` target. It masks no Hull-side uninitialized read. I chose the surgical suppression over instrumenting all of `libkeel` (which would have a far larger blast radius across every suite that links Keel). Happy to switch to full Keel instrumentation if you'd prefer that trade-off.

**Also verified the one remaining log line is benign:** the `js_runtime.c:919 "module evaluation did not settle"` message is expected output from the *passing* negative test `js_extract_liveness.never_settling_await_is_fatal` — it appears on main's green MSan run too, and `test_js` reports PASSED 156.

Pushed 860ae26f. Re-running the exact-head matrix; will report the terminal result.

farkasmark:
## MSan: suppression was impossible; instrumenting Keel instead (b240b4bc)

Per your direction, replaced the suppression with **instrumenting Keel under MSan**.

**Why the suppression had to go:** MemorySanitizer does not honor suppressions files for `use-of-uninitialized-value` (unlike TSan/LSan). Confirmed on 860ae26f: the file loaded at the correct absolute path with no parse error, yet the identical `psx_get_so_error` report fired and MSan exited. A raw `getsockopt` in Hull was also rejected — Keel's own async HTTP client does the same `kl_sock_get_so_error` provider call, so that would diverge from Keel's canonical connect-completion idiom, not improve it.

**The fix, honoring each constraint:**
- The `msan` target now builds Keel with `CC="clang -g -fsanitize=memory,undefined -fno-omit-frame-pointer"` — the same MSan flags Hull applies to every read-path TU it compiles (mbedTLS/QuickJS/Lua/SQLite/sh_arena/…).
- **Keel's compiled vendored components are instrumented too.** Verified by dry-running the compile lines:
  - `clang -g -fsanitize=memory,undefined -fno-omit-frame-pointer … -c -o src/socket_posix.o …`
  - `clang -g -fsanitize=memory,undefined -fno-omit-frame-pointer … -c -o vendor/llhttp/llhttp.o …`
- Keel → mbedTLS is now instrumented → instrumented (Hull compiles the mbedTLS objects under MSan in the same target). WAMR stays the sole uninstrumented substrate (via `HL_MSAN`); Keel does not call it.
- **Removed** `tests/msan_suppressions.txt` and all `MSAN_OPTIONS=suppressions=…` wiring.
- **No raw getsockopt** — the provider path is unchanged.
- **`test_smtp_e2e` and `test_smtp_transport` remain in the MSan run.**
- Any newly exposed Keel uninitialized read will be treated as a real defect to diagnose, not a reason to exclude a suite.

This tracks Keel as the instrumented objects (not merely linking an MSan runtime) — the `socket_posix.o` length operand is now shadow-tracked, so the boundary is eliminated at source.

I can't build MSan locally (Apple clang has no MSan), so the exact-head matrix on b240b4bc is the validation. Re-running the complete matrix now.

farkasmark:
## Exact-head matrix on b240b4bc: fully green ✅

- **MSan + UBSan: pass** — 91/91 tests, **0** `psx_get_so_error` reports, **0** MemorySanitizer warnings, **0** suppressions referenced (file removed). The job log confirms `src/socket_posix.o` compiled with `-fsanitize=memory` — Keel is genuinely instrumented, not just linked against the MSan runtime. Instrumenting Keel exposed **no** latent uninitialized read.
- Cosmopolitan (APE) ✅ · Reproducible Cosmo a / b / compare ✅ · ASan + UBSan ✅ · Static Analysis (cppcheck) ✅ · Lint ✅ · all Build flavors ✅ · applicability-aware gate ✅ · DCO Sign-off ✅
- PR state: `MERGEABLE` / `CLEAN`, 0 non-success checks.

Slice 2b is ready to squash-merge at b240b4bc.

PR #450, opened 2026-09-01
docs(smtp): Slice 2c plan (runtime scheduling) - PLAN ONLY, not implemented

farkasmark:
Docs-only PR: the Slice 2c implementation plan for review. **No implementation** is included and none begins until this plan PR is approved.

`docs/smtp_keel_slice2c_plan.md` moves the SMTP transport from model 1 (synchronous on the event-loop thread) to model 2 (a bounded worker owns the private event context while the runtime suspends), per design section 6.

This v3 incorporates two rounds of plan review. Key frozen decisions:

- **API + fallback (§2):** C sync; Lua yields (same table); JS always returns a `Promise<{ok,error}>`. Synchronous execution happens ONLY with no active event loop; under an active loop the op either submits or returns an existing failure token, never blocks the loop.
- **One linearizable op state machine (§6):** `QUEUED→RUNNING→COMPLETING→DONE` with `CANCEL_REQUESTED` contending on the *same* atomic as `RUNNING→COMPLETING`; the CAS out of `RUNNING` is the linearization point; release/acquire publishes the terminal payload; refcount decides which side frees what.
- **Two-phase teardown (§5):** cancellation request vs confirmed detachment; runtime releases suspension only after a published terminal; transport storage lives until detach/close ACKs; fail-closed `teardown_leaked`.
- **Frozen error mapping (§0/§8/§9):** existing user-visible strings preserved; cancellation/deadline/scheduling failures all map to `connect_failed`, distinctions only in structured audit metadata.
- **Honest deadline (§8):** post-resolution operation deadline (DNS is not bounded; the sandbox forbids the cancellable async resolver).
- **TLS trust material (§11):** an immutable server-retained CA buffer (embedded bundle, or a file read once before sandbox); per-worker-thread `KlTlsCtx` built from it via `*_create_from_buf`; destroyed before the buffer/allocator.
- **SMTP cap + fairness (§9):** cap computed at wiring from the known pool size, atomic queued+running counter; a one-worker pool disables async admission (no false headroom claim); saturation test proves no DB/compute starvation.
- **Shutdown ordering (§10):** explicit order; post-loop-stop completions are discarded (not assumed delivered) — flagged as a must-verify against both async backends.
- **Deferred tests (§12):** injected deterministic address list + attempt-observation seam; Happy-Eyeballs first attempt stays pending; connect-deadline vs operation-deadline isolated so each test proves which timer fired; realistic sanitizer/backend matrix.

Open items parked for implementation review are in §15. Implementation remains stopped pending approval of this plan.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -109,6 +109,7 @@ specific signature, parameter list, or return value.
 | [`reporting_ir_design.md`](reporting_ir_design.md) | Proposed reporting IR - **deferred** (over-scoped as the immediate next feature). |
 | [`sidecar_design.md`](sidecar_design.md) | Native sidecars - design only (Phase 0), not yet scheduled. |
 | [`smtp_keel_client_design.md`](smtp_keel_client_design.md) | Design-only transition of Hull SMTP from direct POSIX/blocking transport to an SMTP-owned composition of public Keel v3 primitives, preserving Hull policy and protocol ownership. |
+| [`smtp_keel_slice2c_plan.md`](smtp_keel_slice2c_plan.md) | Plan-only architectural gate for SMTP Slice 2c: moving the transport onto a bounded worker (model 2), with frozen decisions for worker ownership, the linearizable op state machine, the post-resolution deadline, the SMTP pool cap, verified backend-shutdown ownership, TLS trust-material ownership, and the zero-Keel link seam. |
 
 ### Roadmap
 
--- a/docs/smtp_keel_slice2c_plan.md
+++ b/docs/smtp_keel_slice2c_plan.md
@@ -2,139 +2,102 @@
 
 Status: PLAN ONLY. No implementation until this PR is approved.
 Branch: `feat/smtp-keel-slice-2c` off `main` @ 76bc29bf (Slice 2b merged).
-Revision: v3 (freezes fallback, the op state machine, scheduling-failure tokens,
-TLS trust-material ownership, the SMTP cap, and shutdown ordering).
+Revision: v4 (freezes all five architectural-gate items: cap formula, deadline
+relationship, audit schema, verified backend-shutdown ownership, and the exact
+zero-Keel link seam).
 
-## 0. Scope (fixed by review)
+## 0. Scope and frozen invariants
 
-Four workstreams, nothing else:
+Four workstreams: (1) runtime/worker scheduling; (2) off-thread resolution; (3)
+the post-resolution operation deadline; (4) deterministic coverage for the four
+deferred behaviors. The reviewed Slice 2b protocol, TLS, parser, teardown, and
+audit contracts are preserved UNCHANGED except the runtime API shape (section 2)
+and the audit/execution split (section 4). PostgreSQL / MySQL are later
+consumers. Native-Windows IOCP is a non-goal; "Windows" is the Cosmopolitan APE
+(poll).
 
-1. runtime/worker scheduling (move SMTP execution off the event-loop thread);
-2. off-thread name resolution;
-3. the post-resolution operation deadline (section 8);
-4. deterministic coverage for the four deferred behaviors.
-
-The reviewed SMTP protocol, TLS, parser, teardown, and audit contracts from
-Slice 2b are preserved UNCHANGED except the runtime API shape (section 2) and the
-audit/execution split (section 4). PostgreSQL / MySQL are later independent
-consumers. Native-Windows IOCP stays a non-goal; "Windows" is the shipped
-Cosmopolitan APE (readiness/poll).
-
-FROZEN PUBLIC ERROR MAPPING: existing user-visible strings are preserved
-verbatim. Cancellation, deadline, and every scheduling failure map to EXISTING
-tokens (sections 2, 8, 9); their distinctions live only in structured audit
-metadata. No Keel string is ever leaked as a user-visible token.
+FROZEN: existing user-visible error strings are preserved verbatim; cancellation,
+deadline, and scheduling failures all map to `connect_failed`, with distinctions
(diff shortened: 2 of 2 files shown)


farkasmark:
## v4: all five architectural-gate items frozen (7299c7b2)

Docs-only amendments; implementation still stopped. Each item is now a frozen value or a test-gated requirement, not an assumption.

**1. SMTP cap formula (§9)** — frozen: `W<=1` disables async SMTP; `W>=2` → `smtp_max_inflight = max(1, floor(W/2))`. Cap computed at wiring (the pool doesn't expose W at runtime); SMTP counts its own queued+running jobs. Tests parameterize W∈{1,2,3,4,5,8} (expect {disabled,1,1,2,2,4}) and prove ≥1 worker always stays outside SMTP admission; a saturation test proves db.async/compute still completes under an SMTP flood.

**2. Deadline relationship (§8)** — frozen: `Dop = monotonic_now_after_resolution + configured_timeout`; `Dstage = min(Dop, now + stage_budget)`; the connect timer gets `Dop - now` (never a fresh full timeout); graceful close bounded by `min(Dop, now + SMTP_CLOSE_GRACE_MS)`; no stage/retry extends `Dop`. Honestly post-resolution (DNS excluded, since the sandbox forbids the cancellable async resolver).

**3. Audit schema (§3)** — frozen fields, emitted only when applicable, public error stays `connect_failed`, pinned by exact-JSON tests: `"schedule": pool_unavailable|cap_reached|queue_full|suspend_failed`; `"terminal": cancelled|post_resolution_deadline`; `"teardown": leaked` (preserved).

**4. Backend-shutdown ownership (§10) — investigated both, not assumed:**
  - **Keel** `kl_thread_pool_free` (thread_pool.c:240): joins, removes watcher, then **drains the done queue calling `done_fn`**, then `cancel_fn` for never-started.
  - **Poll** `poll_pool_free` (poll.c:702): joins, fires `cancel_fn` for never-started, but **does not drain the completion queue → a `done_fn` finishing during shutdown is dropped** (matching the OOM-drop at poll.c:618).
  - Neither `free()` fires `on_cancel` for a suspended op ("graceful shutdown is the caller's job").
  
  This is a real divergence. Design consequence (frozen): `done_fn` is resume-only and may be dropped — it never owns a free/ref-drop; `work_fn` (always runs, since `pool_free` joins) owns transport teardown + storage free + worker-ref drop on both backends; the runtime ref is dropped by exactly one of {on_resume, per-request cancel, the server-shutdown SMTP cancel sweep}, arbitrated by the refcount, with the sweep guaranteeing it since `free()` won't. Added the three-case ownership table + a **backend-parity test requirement** (shutdown-with-op-in-flight runs under both backends, asserting no leak/UAF/exactly-once in every case).

**5. Exact link seam (§12)** — named: WEAK `hl_async_backend()`/`hl_async_backend_poll` in `src/hull/async/poll.c` (0 `kl_` refs, what a Keel-less base links); STRONG `hl_async_backend_keel` anchor in `src/hull/async/keel.c` (22 `kl_` refs, sole `kl_thread_pool_*` referencer), composed via `libhull_feature-keel.a`. The SMTP submit/resume glue (`worker_smtp.o`, mirroring `worker_db.o`) references ONLY the `HlAsyncBackend` vtable via `hl_async_backend()`, never `kl_` directly. Compute-only can't pull Keel because (a) SMTP is only composed when `hull/smtp`/`hull/email` is declared, and (b) the glue's weak default is the Keel-free poll backend. Test: the existing `nm app | grep kl_thread_pool` → empty, PLUS a negative test that `nm -u worker_smtp.o` has no `kl_` and that a deliberate direct `kl_thread_pool_submit` bypass makes it fail.

Gates rerun green (em-dash, milestone-narration, docs-integrity; catalogued in docs/README.md). Returning for final architectural approval; no SMTP production code will move until this PR is approved.

farkasmark:
## Factual correction round (4d984ece)

All three record inconsistencies + the wording fix are corrected; docs-only, gates rerun green.

**1. Link seam names the right suspension path (§12).** Now documents TWO seams: **Seam A** worker execution via `hl_async_backend()` (weak poll `async/poll.c` / strong keel `async/keel.c`, `pool_submit` only); **Seam B** active-request suspension/resumption via `hl_net_op_suspend()` / `hl_net_op_complete()` on an `HlAsyncCtx` bound to the request `net_ctx` (the db/compute pattern, exactly `mod_db.c:787`), weak stubs vs strong `src/hull/net/keel.c` — **not** the `HlAsyncBackend` vtable. The nm requirement is updated: `worker_smtp.o` must have **no undefined `kl_` symbols** while **permitting** the two `hl_net_op_*` symbols.

**2. Shutdown order built around the real primitive (§10).** Rewritten around `pool_free()` as the single call that signals/joins/dispatches-or-drops/cancels/destroys (no separate "join now, destroy later"): stop submissions → mark shutdown + request cancellation → `pool_free()` → verify every worker ref is gone → release retained runtime refs → free CA material + allocator. Added that per-worker pthread-TLS destructors run **as workers exit during `pool_free()`**, so the allocator must stay alive **through** that call.

**3. Runtime-ref ownership made consistent (§5 ↔ §10).** One frozen contract: a per-request cancel marks the continuation **non-resumable** but **retains** its runtime-side ref until terminal publication (or the post-`pool_free()` sweep); it never releases the last runtime-side ownership before the worker publishes terminal. Removed the contradictory "per-request cancel drops the runtime ref" wording.

**Wording fix (§10/§11):** a `KlTlsCtx` needs its **allocator** at destroy, **not** the CA buffer (mbedTLS copies the certs in at creation). Retaining the immutable CA buffer through pool teardown is stated as a reasonable server-lifetime policy, no longer justified by a false destroy-time dependency.

No SMTP production code has moved. Returning #450 for final architectural approval.

PR #460, opened 2026-09-06
ci: build Hull from source on Windows under MSYS2

farkasmark:
## What

`windows-source-build` failed on its first dispatch before compiling a single translation unit. It is now **green**: [run 34038376368](https://github.com/artalis-io/hull/actions/runs/34038376368) builds Hull from source on Windows in 2m33s and produces a 17.6 MB APE that runs and reports Windows-correct facts.

Getting there uncovered four distinct problems, each only visible once the one before it was fixed.

## 1. Git Bash + Chocolatey make cannot build Hull at all

The original job assumed `choco install make` plus Git Bash would work. It cannot, and the reason is structural rather than a configuration slip.

A **native Win32** make does not exec `sh` with an argv — it builds a *Windows command line*, which MSYS `sh` re-parses under Windows rules. That destroys the POSIX `\"` escaping Hull's Makefile uses for its six version macros (`HL_VERSION`, `HULL_VENDOR_KEEL_COMMIT`, …). Measured:

| Invocation | argv |
|---|---|
| native make, recipe **with** `\"` | `argc=2` — everything after the first `\"` collapses into one argument |
| native make, recipe **without** `\"` | `argc=7` |
| identical line via `sh -c` | `argc=8`, `-DHL_VERSION="0860e32"` intact |

Confirmed on the runner ([run 34025202379](https://github.com/artalis-io/hull/actions/runs/34025202379)) for **both** native makes present there — mingw64's and Chocolatey's — which mangle identically. The recipe and the shell are both fine; the Windows command-line hop is what breaks it, so no PATH or shell fix reaches it.

**Fix:** `msys2/setup-msys2`, pinned to `66cd2cce` (v2.32.0), installing `make binutils vim diffutils git` — `binutils` for `ar`/`nm`, `vim` for `xxd` (~54 Makefile call sites), `git` so the build stamps a real revision instead of falling back to `"dev"`.

## 2. `make` was never Chocolatey's anyway

cosmocc's bundle ships its own `make`, and the workflow prepended `cosmo/bin` to PATH — so `make` resolved to that APE in every early run. It also **loses its exit status** under a Windows shell: it printed `*** Error 127` and still returned 0, which made an early probe report success over a failed build. `cosmo/bin` is now appended, and the probe refuses to proceed if `make` resolves there.

## 3. `HL_OPT=-O0` never reached Keel

`mk/vendor/keel.mk` hands its sub-make `CC`, `AR` and the TLS/compress config and nothing else, so Keel built at its own hardcoded `-O2`. [Run 34035825889](https://github.com/artalis-io/hull/actions/runs/34035825889) died there: 427 TUs at `-O0`, 81 at `-O2`, then **41 minutes of total silence** with two `-O2` miniz TUs in flight until the 45-minute cap — the cosmocc wedge this file's header documents. The job had never actually been running the configuration it claimed.

Keel exposes no opt knob and no append hook, and its `CFLAGS`/`VENDOR_CFLAGS` are plain `=`, so the only lever short of restating all of Keel's flags is the compiler: gcc honours the **last** `-O`. A shim appending ours wins. It is named `cosmocc` exactly, because Keel sets `COSMO_FAT` (the aarch64 half of the fat APE) only for `ifeq ($(CC),cosmocc)`.

## 4. Two Hull-side defects found, deliberately NOT fixed here

Both need a Keel-side knob, so they are recorded in the workflow header rather than worked around further:

1. **`HL_OPT` should propagate to the Keel sub-make** instead of being forced by a CI-local shim.
2. **`KEEL_EXTRA_CFLAGS` / `KEEL_EXTRA_LDFLAGS` are not referenced anywhere in Keel's tree.** The LTO and CFI flags `mk/vendor/keel.mk:34` passes are silently discarded — on **every** platform, not just Windows. Worth looking at independently of this job.

## Scope of the MSYS2 requirement

Stated precisely in `docs/windows_install.md` and `README.md`, because it is easy to overstate in either direction:

- MSYS2 is required **only** to build Hull *itself* from source on Windows.
- **Installing** Hull (`install.ps1`) and **building apps** with the released `hull.com` need no MSYS2, no Visual Studio and no WSL — `hull build` is compiler-free by default and its cosmo path is `hull tools install cosmocc`.

## Also

- The probe now asserts rather than assumes: an MSYS-native make, escaped quotes surviving to argv, `/bin/sh` routing, cosmocc's POSIX `$0`, a full tool inventory, and marker-based (not exit-status-based) verification — so a make that loses its exit status cannot report success. Two earlier fixes, a `<drive>\bin` junction and a cosmocc `$0` wrapper, proved unnecessary under MSYS2 and were removed; the probe verifies the properties they provided.
- Keel's compile lines still **echo** `-O2` because make echoes the recipe and the shim appends `-O0` at exec time. The header says so explicitly — grepping the log for `-O2` is not evidence the shim failed.
- Trigger stays `workflow_dispatch`. Now that it is green, switching to the path-filtered `pull_request` trigger (already drafted in the header) is a reasonable follow-up.

https://claude.ai/code/session_013crBdW6cEF5Q2xuASxeivL

### DIFF
--- a/.github/scripts/dump-stuck-build.ps1
+++ b/.github/scripts/dump-stuck-build.ps1
@@ -1,0 +1,88 @@
+# Dump the processes of a wedged cosmocc-on-Windows build.
+#
+# WHY POWERSHELL: only the Win32 process table exposes a process's full command
+# line and cumulative CPU time. An MSYS `ps` shows neither usefully, and the one
+# question worth answering about this hang is whether the stuck compiler is
+# BURNING CPU or BLOCKED - which needs CPU sampled twice.
+#
+# Two snapshots `IntervalSeconds` apart, reporting the delta:
+#   dcpu ~= 0        -> blocked (file lock, pipe, temp-file wait). Suspect the
+#                       environment: cosmocc's mktemper temp files, AV scanning,
+#                       a stalled pipe - not the optimizer.
+#   dcpu ~= interval -> spinning. Suspect codegen/optimizer on that exact TU.
+#
+# Context: .github/workflows/windows-source-build.yml wedges intermittently
+# (~1 run in 4). Two captured signatures so far, and they do NOT share a cause:
+# miniz at -O2 (fixed - Keel never got HL_OPT), and stdlib_js_registry at -O0
+# (open). Both showed ~0 CPU indefinitely, which is what this script exists to
+# confirm or refute with actual numbers.
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+param([int]$IntervalSeconds = 5)
+
+$ErrorActionPreference = 'Continue'
+
+# Match on the WHOLE base name, never a substring. An unanchored alternation
+# here is worse than useless: `as`, `ar` and `sh` match 1Password.exe,
+# StartMenuExperienceHost.exe and ShellHost.exe, burying the one process the
+# script exists to show. Anchored set + a cosmo-toolchain catch-all instead.
+$exactNames = @(
+    'cc1', 'cc1plus', 'gcc', 'g++', 'clang', 'make', 'sh', 'bash', 'dash',
+    'ld', 'ld.bfd', 'as', 'ar', 'nm', 'xxd', 'ape', 'apelink', 'fixupobj',
+    'mktemper', 'objbincopy', 'zipobj', 'assimilate'
+)
+
+function Get-BuildProcs {
+    Get-CimInstance Win32_Process -ErrorAction SilentlyContinue |
+        Where-Object {
+            $base = $_.Name -replace '\.(exe|com|ape)$', ''
+            ($exactNames -contains $base) -or ($base -like '*cosmo*')
+        } |
+        ForEach-Object {
+            $p = Get-Process -Id $_.ProcessId -ErrorAction SilentlyContinue
+            [PSCustomObject]@{
+                ProcId = $_.ProcessId
+                Parent = $_.ParentProcessId
+                Name   = $_.Name
+                Cpu    = if ($p) { [math]::Round($p.CPU, 2) } else { $null }
+                RssMb  = if ($p) { [math]::Round($p.WorkingSet64 / 1MB, 1) } else { $null }
+                Cmd    = $_.CommandLine
+            }
+        }
+}
+
+Write-Host "=== stuck-build diagnostics ==="
+Write-Host ("sampling process CPU {0}s apart" -f $IntervalSeconds)
+Write-Host ""
+
+$first = @(Get-BuildProcs)
+Start-Sleep -Seconds $IntervalSeconds
+$second = @(Get-BuildProcs)
+
+if ($second.Count -eq 0) {
+    Write-Host "no matching build processes found - the wedge may not be a live process"
+    exit 0
+}
+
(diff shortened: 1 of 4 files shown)


farkasmark:
Two updates since the review request:

- **Path-filtered `pull_request` trigger is now enabled** (`0219f13d`), the job having passed end to end in [run 34038376368](https://github.com/artalis-io/hull/actions/runs/34038376368). The drafted path list went in as written; `workflow_dispatch` is kept alongside it. The list includes this workflow file, so **this PR exercises the new trigger on itself** — the Windows job should appear in its own checks.
- **Follow-up issue #461** covers both Keel flag-propagation problems together: add supported hooks in Keel, wire Hull to them, and verify the actual compile/link commands for `HL_OPT`, LTO and CFI. The `-O0` shim in this workflow stays until that lands, and both the header and the shim now say so and name the issue, so the workaround gets deleted rather than inherited.

Severity note recorded in #461: `HL_ENABLE_LTO` / `HL_ENABLE_CFI` are opt-in and set by no workflow or the release pipeline, so **shipped binaries are unaffected**. The defect is that `make HL_ENABLE_CFI=1` reports CFI enabled while Keel's TUs are silently uninstrumented.

farkasmark:
Fixed in `f91dcebf` — the finding was correct, and there was a third gap alongside the two reported.

**What was wrong.** `docs/windows_install.md` advertised `make CC=cosmocc HL_OPT=-O0`, which is not the configuration CI validates:

- **No `-O0` shim.** `HL_OPT` does not reach Keel's sub-make (#461), so Keel builds at its own hardcoded `-O2` and wedges on miniz — *deterministically*. Anyone following the doc hit exactly the failure [run 34035825889](https://github.com/artalis-io/hull/actions/runs/34035825889) hit.
- **`HL_ENABLE_WASM` left on**, while the workflow sets it to `0`. WAMR under cosmocc-on-Windows is unproven, so the documented build was both broken *and* wider than anything tested.
- **Third gap, not reported, my omission:** the `pacman` line was missing `diffutils` and `git`. I added those to the workflow after the first MSYS2 run reported them, and never back-ported them here.

**What it says now.** The block mirrors the workflow exactly — package set, SHA-verified cosmocc at the same pin, the shim, PATH order (`cosmo/wrap` first, `cosmo/bin` last so its bundled `make` cannot shadow MSYS2's), and `HL_ENABLE_WASM=0` — and states which flags are load-bearing and why rather than presenting them as taste.

It also now warns that the build wedges **~1 run in 3** (#462). Omitting that was arguably the worse half: a reader whose build goes silent should know it is a known open defect and that a retry is cheap, not conclude their own setup is broken.

**Verification.** The doc's shim recipe was run verbatim against a stub — a Keel-style `-O2 …` call comes out `-O0`. The cosmocc version and SHA-256 are checked to match the workflow's pin. Two accuracy fixes to my own text while there: the fetch claimed to be "SHA-pinned" while verifying nothing (it now runs `sha256sum -c`), and the intro no longer implies the command is validated beyond "this is what CI runs".

Checked for the same stale recipe elsewhere: the other `CC=cosmocc` references (CLAUDE.md, `docs/agent_guide.md`) are Linux/macOS cosmo builds, which are the supported release path and do not hit the Windows-specific hang. The README anchor still resolves — the heading is unchanged.

PR #473, opened 2026-09-08
ci: layer-cache the reproducible build container (needs rework: evicts the repro artifacts)

farkasmark:
Options (b) and (d) from the outage discussion on #472. Neither disturbs the
registry-less design; (a), publishing the image to GHCR, is left alone.

## The problem

The reproducible build container pins its apt archive to a point in time on
`snapshot.ubuntu.com`. That is a single Canonical service with **no public
mirror** - ordinary Ubuntu mirrors carry the *current* archive and have no
`/ubuntu/<timestamp>/` URL space at all - so while it is degraded the image
cannot be built anywhere, and there is no host to fail over to. DNS already
returns three IPs (`185.125.189.36/.37/.69`).

On 2026-09-08 it returned 502/503 across all three for hours. Four jobs failed
on every push, inside `docker build`, before a line of Hull was compiled:

```
E: Failed to fetch http://snapshot.ubuntu.com/.../Packages  503  Service Unavailable
ERROR: failed to solve: ... apt-get install ... exit code: 100
```

## (d) Retries

`apt-get -o Acquire::Retries=5`, and `--retry 5 --retry-all-errors` on the
cosmocc fetch. This absorbs a blip. It does **not** absorb an outage that long,
and the comment says so rather than implying the problem is solved.

## (b) Layer cache, opt-in

The three CI reproducibility jobs opt into a buildx `type=gha` cache via a new
`cache` input on `hull-build-container`; a hit skips the network entirely.

**Caching does not weaken what those jobs prove.** The image is a pure function
of the Dockerfile - the base is digest-pinned, and the apt snapshot is derived
from that base - so a cache hit and a cache miss produce the same layers by
construction. That is precisely the property these jobs assert. A hit only skips
the fetch. The scope is per runner arch, since amd64 and arm64 build genuinely
different images and one scope would make them evict each other.

**Release deliberately does not opt in**, which is why this is an input rather
than unconditional. The same action is used in seven places in `release.yml`,
and release provenance should keep resting on the digest and snapshot pins alone
- no mutable cache, and no `docker/setup-buildx-action` /
`docker/build-push-action` in that path. CI trades a little trust surface for the
ability to run during someone else's outage; a release does not need that trade,
because it is a deliberate act that can simply be retried.

Both new actions are SHA-pinned, resolved from their upstream tags rather than
transcribed by hand.

## What this does not do

**It does not unblock anything today.** The first build after this change is a
cache miss, so it still needs the mirror. The claim is only that the *next*
outage does not take CI down.

Worth knowing about cost: the image carries clang plus a ~330 MB cosmocc, so
each arch's cache runs to roughly a gigabyte against the repo's 10 GB budget and
can evict other caches.

`docs/security.md` ("Build-environment immutability") and
`.github/docker/README.md` record the availability caveat and the CI/release
asymmetry, since both previously described the registry-less design as though
availability followed from it.

### DIFF
--- a/.github/actions/hull-build-container/action.yml
+++ b/.github/actions/hull-build-container/action.yml
@@ -17,6 +17,37 @@ inputs:
       digest-pinned base image, which is already deterministic.
     required: false
     default: ""
+  cache:
+    description: >
+      Cache the image layers in the GitHub Actions cache (buildx type=gha).
+
+      OFF by default, and deliberately so. This action also builds the image
+      the RELEASE pipeline uses, and a cached layer is a different trust anchor
+      from the base-image digest + apt snapshot pins the Dockerfile documents:
+      it introduces a mutable store, plus two Docker-org actions, into a path
+      that today depends on neither. Release therefore keeps building from
+      source every time.
+
+      CI's reproducibility jobs opt in, because the alternative is worse for
+      them. snapshot.ubuntu.com is a single Canonical service with no public
+      mirror - ordinary Ubuntu mirrors carry the CURRENT archive and have no
+      /ubuntu/<timestamp>/ space at all - so when it degrades, every one of
+      those jobs fails inside `docker build`, before a line of Hull is
+      compiled. Measured 2026-09-08: 502/503 across all three of its IPs for
+      hours, four jobs red on every push, and `apt-get -o Acquire::Retries`
+      (added in the Dockerfile) does not help against an outage that long.
+
+      Caching does not weaken what those jobs prove. The image is a pure
+      function of the Dockerfile: the base is digest-pinned and the apt
+      snapshot is derived from that base, so a cache HIT and a cache MISS
+      produce the same layers by construction - that is exactly the property
+      the reproducible build asserts. A hit simply skips the network.
+
+      Cost worth knowing: the image carries clang plus a ~330 MB cosmocc, so
+      each arch's cache runs to roughly a gigabyte against the repo's 10 GB
+      budget, and can evict other caches.
+    required: false
+    default: "false"
   user:
     description: >
       Container user. Default (empty) runs as root, fine when the job only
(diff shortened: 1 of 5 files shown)


farkasmark:
Splitting this: option **(d)** (the apt / cosmocc retries) is now #496 and can land on its own. This PR keeps option **(b)**, the buildx `type=gha` layer cache, which needs rework before it can merge.

## Why (b) cannot land as-is

The `Reproducible build (Cosmo compare)` job fails here, and the cause is this PR rather than infrastructure noise:

| | Reproducible build (Cosmo compare) |
|---|---|
| `main` — runs `34744612348`, `34741943136`, `34710997642`, `34710981765` | success 4/4 |
| this PR — `34744625089`, `34746621541` | **fail**, identical error both times |

Both failures are:

```
Unable to download artifact(s): Unable to download and extract artifact:
Artifact download failed after 5 retries.
FAIL: reproducibility-cosmo-compare: failure
```

The compare step never runs, so nothing was found to differ — the job dies fetching `hull-cosmo-repro-a` / `-b`. (Both are listed at an identical 7,231,081 bytes, which is what a reproducible pair should look like.)

The mechanism is the one this PR's own `cache` input documents:

> each arch's cache runs to roughly a gigabyte against the repo's 10 GB budget, and can evict other caches

`type=gha` and Actions artifacts share that budget and that backend. The jobs that can no longer retrieve their artifacts are exactly the jobs that just began writing ~1 GB of layers per arch, and it does not happen on `main`. Merging as-is would leave the bit-identical-build gate red on `main`.

## Possible directions

- Do not opt the **reproducibility** jobs into the cache — they are the ones with artifacts to lose, and the outage protection may be worth more on the other jobs.
- Narrow the cache scope, or `mode=min` rather than `mode=max`, so it stops crowding the budget.
- Keep the cache but move the repro artifacts off the Actions artifact store.

Confirmed separately: the failure survives a refresh against current `main` (the second run above was on `80c7f8ae`, after this branch was brought up to date), so it is not the stale 2026-09-08 outage.

PR #478, opened 2026-09-09
ci(windows): enforce test_fs once the mmap-window gaps are fixed

farkasmark:
Stacked and **not mergeable yet** - opened so it is ready rather than remembered.

Moves `test_fs` from #475's KNOWN tier to ENFORCED, and drops the baseline that quietly permitted 3 failures.

## Order this has to land in

1. #477 - the windowed-mmap fixes (64 KiB allocation granularity, no view past EOF, and the symlink test that asserted nothing)
2. #475 - the workflow this edits
3. this

Landing it before #477 turns the nightly red: the branch would enforce a suite whose fix is not present.

Base is `ci/windows-test-and-app-build` (#475's branch) rather than `main`, so the diff here is the single commit rather than all of #475.

## Verified

Ran the modified step's own logic against binaries built from #477:

```
===================== ENFORCED =====================
ok  test_host  test_tool  test_vfs  test_static  test_cacert
ok  test_dispatch  test_csp  test_parse_size  test_signature
ok  test_release  test_fs
====================== KNOWN =======================
known  test_compiler: 5 (baseline 5)
Windows unit suites: enforced clean, known failures at baseline.   (exit 0)
```

## One thing to actually check rather than assume

That verification is a **developer box**. The runner and a dev box already disagreed by one on this suite (3 failures vs 4), so the first nightly after this lands is worth reading. If the runner still shows a failure, the ratchet names the case - which is what it is for.

### DIFF
--- a/.github/scripts/make-watched.sh
+++ b/.github/scripts/make-watched.sh
@@ -1,0 +1,111 @@
+#!/usr/bin/env bash
+#
+# Run ONE make invocation under a stall watchdog, retrying ONLY a stall.
+#
+# WHY THIS EXISTS. cosmocc's cc1 blocks indefinitely at ~0 CPU on roughly 12%
+# of large-translation-unit compiles on Windows (hull#462). That is a
+# cosmocc-on-Windows bug, not a Hull one, and it is why every make invocation
+# in windows-source-build.yml needs a watchdog rather than just a job timeout:
+# a wedge otherwise burns the full cap and reports nothing useful.
+#
+# WHY A SCRIPT AND NOT INLINE BASH. This logic was inline in the "Build hull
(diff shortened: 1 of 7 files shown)


farkasmark:
Superseded by #475, which now carries this change directly.

This was opened as a stacked draft because #477 had not landed yet. It has, so the dependency is gone - and leaving the promotion out of #475 would actively break the first nightly: with the mmap fixes in main, `test_fs` returns 0 failures against a baseline of 3, and the ratchet fails on improvement by design.

Folded into #475 rather than merged separately, so there is no window where main enforces a stale baseline. Verified on the runner in #475's validation: 11 enforced suites clean including `test_fs`.

Issue #502, opened 2026-09-14
Windows: manifest extraction reports success for an app that throws before app.manifest


farkasmark:
## Summary

On Windows (cosmo APE), `hull build` of an app whose top level throws **before** `app.manifest` proceeds as though extraction succeeded. No `manifest extraction failed` diagnostic is emitted and the build continues into codegen.

On POSIX the same app aborts the build, which is the documented contract — `tests/e2e_modular_resolution.sh` exists to enforce exactly this:

> a manifest-extraction failure that controls composition is FATAL (build aborts, no binary), not a silent nil manifest

## Evidence

App under test (`$brk/app.lua`):

```lua
error("cannot reach app.manifest")
app.manifest({ modules = {} })
```

`hull build` on Windows, captured with `2>&1`:

```
hull build: 1 Lua file(s) from /D/a/_temp/tmp.iii5qRjAn2/broken
hull build: emitting app_registry.o (elf/x86_64, 1 entries)...
hull build --no-compiler: this hull has no bundled app_main.o
```

With `--verbose`, identical apart from the sandbox line:

```
INFO  src/hull/sandbox_tool.c:300: [sandbox] tool mode applied (20 unveiled paths)
hull build: 1 Lua file(s) from /D/a/_temp/tmp.iii5qRjAn2/broken
hull build: emitting app_registry.o (elf/x86_64, 1 entries)...
hull build --no-compiler: this hull has no bundled app_main.o
```

`emitting app_registry.o` is the key line: the build did not stop at extraction. It ran to the link and only halted there, for an unrelated reason (this CI hull has no bundled `app_main.o`). Had it been an `EMBED_PLATFORM` build, it would have produced a binary.

Expected (and what POSIX does) is `extract_manifest_or_fatal` in `stdlib/cli/lua/hull/feature_compose.lua`:

```
hull build: manifest extraction failed - it controls feature composition,
so the build cannot continue.
```

## Why this matters beyond one error message

`extract_manifest_or_fatal` is shared by `hull build`, `hull deploy` and `hull eject`. If extraction silently reports success on Windows, feature composition is decided from a manifest that was never read — the "silent nil manifest" case the fatal path was added to prevent. The manifest is also the capability declaration, so the composed result is not merely wrong, it is unreviewed.

[...]

PR #504, opened 2026-09-14
test(windows): skip the fatal-extraction assertions on a hull that cannot link

farkasmark:
Closes the test side of #502 — which turned out **not** to be a Hull defect.

## What the instrumentation showed

Probing `M.extract_manifest` at entry and every exit on Windows ([run 34858796974](https://github.com/artalis-io/hull/actions/runs/34858796974)):

**Where hull can link** (app-build job):

```
hull build: 1 Lua file(s) from .../demo
[diag502] ENTER .../demo
[diag502] AFTER-LOADFILE true load_err=nil intends=true
[diag502] RETURN table: 0x... err=nil
```

**Where it cannot** (the e2e job) — no `ENTER` for *any* app, including the valid ones:

```
hull build: 4 Lua file(s) from .../mod_lua      <- no ENTER
hull build: 4 JS file(s) from .../mod_js        <- no ENTER
hull build: 1 Lua file(s) from .../broken       <- no ENTER
```

That hull has no bundled `app_main.o`, so `hull build` fails at that check **before** manifest extraction is attempted. The absent `manifest extraction failed` message is a consequence of the build stopping earlier, not of extraction wrongly reporting success.

The assertion was therefore asserting on a code path that had not run.

## The change

It now carries the same unsupported-configuration guard the build assertions in `run_lang` already had — that guard postdates this check, which is why this one never got it.

Verified the guard fires in the right direction rather than trusting the pattern:

| output | verdict |
|---|---|
| `no bundled app_main.o` | SKIP |
| `cannot find libhull_platform.a` | SKIP |
| `manifest extraction failed - it controls …` | ASSERT |
| `hull build: 1 Lua file(s) from /tmp/x` | ASSERT |

So a genuine extraction failure, and silent wrong behaviour, both still fail the suite.

## Also

Drops a comment describing the refuted theory (extraction reporting success, composition decided from a manifest never read). It is not happening, and leaving it in would send the next reader after a defect that does not exist.


farkasmark:
Green on Windows: [run 34930891987](https://github.com/artalis-io/hull/actions/runs/34930891987) probes `e2e-modular-resolution: ok (24s)`, against `FAILED` on every run of this suite before it.

The last one standing was the loader-recovery boundary, fixed in 738e1cd6 — same `if "$HULL" build` defect as its sibling, just the next in line. Full probe output:

```
PASS: lua dev resolves routes -> ./../lib -> ./../models
SKIP: lua build assertions (this hull cannot link an app here)
PASS: js dev resolves routes -> ./../lib -> ./../models
SKIP: js build assertions (this hull cannot link an app here)
PASS: escape-past-root fails closed in build (no binary)
...
PASS: JS eval/Function stay removed after importing a helper
e2e_modular_resolution: ALL PASS
```

One thing the skips expose, worth deciding separately rather than folding in here: on a hull that cannot link an app, the two `hull_rc`-based fail-closed checks (escape-past-root, JS pre-manifest throw) still *pass* — but they would pass whatever hull did, since that build fails before extraction either way. By the argument this branch has been applying, those are vacuous passes and should be skips too. They are not failing, so I have left them alone.

Issue #427, opened 2026-08-27
QuickJS teardown-GC abort on self-referential-microtask app during manifest extraction


farkasmark:
## Summary

Tearing down a QuickJS runtime that ran a **self-referential-closure microtask churn** intermittently **SIGABRTs in QuickJS's teardown GC** (`__JS_FreeValueRT` double-free of a promise resolver `js_c_function_data`, reached via `JS_RunGC` ← `JS_FreeRuntime`). This surfaces during JS manifest extraction (`hull build` of a `.js` app) when the app's top-level schedules an endlessly-requeuing microtask via a module-local closure.

This is **not** an authorization bypass, a hang, or a successful bad build: the loader correctly bounds the drain, fails closed (PENDING → fatal), and produces **no binary**. The only defect is that the process aborts in teardown instead of exiting cleanly. Discovered while hardening the extraction settle path in #426.

## Minimal reproducer

```js
// app.js
import { app } from "hull:app";
const p = new Promise(() => {});                 // never resolves
function spin(){ Promise.resolve().then(spin); } // module-local const -> closure garbage cycle
spin();
await p;                                          // holds the module PENDING
app.manifest({ modules: [] });                    // unreachable
```

```sh
hull build ./app_dir --no-verify-platform   # ~50% of runs exit 134 (SIGABRT); else exit 1 (clean fatal)
```

## Observed behavior

- **~50% SIGABRT** (exit 134), rest clean fatal (exit 1). No binary in either case.
- Backtrace: `abort` ← `__JS_FreeValueRT` ← `js_c_function_data_finalizer` ← `free_gc_object` ← `JS_RunGC` ← `JS_FreeRuntime` ← `hl_js_free` ← `hl_manifest_extract_js_from_file`.
- **Orthogonal to loader liveness / draining:** fires even at a drain cap of **0** (zero jobs executed) ~50%. The corruption is inherent to the self-referential closure cycle existing at teardown, not to how many jobs run.
- **Trigger is the garbage cycle, not microtask volume:** a never-settling `await` alone (0 jobs) is 0-abort; a *non*-self-referential finite promise burst is 0-abort; a **global-rooted** rescheduler (`globalThis.__f = () => Promise.resolve().then(globalThis.__f)`, not a garbage cycle) is **0/60 stable under load** despite being a genuine endless queue.

## What did NOT help (all made it worse — they run/churn more)

- Interrupt-poisoning the drain to force-empty the queue.
[...]

PR #505, opened 2026-09-15
test(windows): e2e_manifest_extract_isolation was asserting on a status it could not read

farkasmark:
Next suite in the Windows e2e migration. Probed verdict was `FAILED`, on:

```
FAIL: run 1 unexpectedly SUCCEEDED (a PENDING module must fail extraction)
```

It did not succeed. The build failed exactly as intended — the shell could not see it. On Windows an APE reports its status shifted left by 8 and a POSIX shell keeps the low byte (jart/cosmopolitan#1521), so `wait` returns 0 for every outcome. Same defect #503 fixed elsewhere, reaching this suite through `wait` rather than `$?`.

**Every** status-dependent check here was affected, not only the one that reported. They are not all the same problem, so they are not all treated the same way:

| check | needs | treatment |
|---|---|---|
| 2a–2d, child protocol | a real status; links nothing | routed through `hull_rc` — still asserted on every host |
| 1, the #427 reproducer | a real status, but is backgrounded for the hang guard | contracts split (below) |
| 3, 4, ordinary builds | a hull that can link an app | skipped where it cannot, with the reason named |

### The split in check 1

Routing the reproducer through `hull_rc` would mean giving up the hang guard that watches it — a hang is a *different* regression and must not read as a pass. So its two contracts are separated:

- **terminates, and leaves no binary** — status-free, holds everywhere, asserted unconditionally.
- **died cleanly rather than on a signal** (the actual #427 contract) — needs a real status, skipped where the shell cannot supply one.

Keeping the first half unconditional is what stops the skip from hiding anything: a build that wrongly *succeeded* still produces a binary, and that assertion still catches it on every host.

### Unchanged elsewhere

`hull_rc_init` no-ops off Windows, so Linux and macOS take the direct call and assert exactly what they did before.


farkasmark:
Green on Windows: [run 34931461950](https://github.com/artalis-io/hull/actions/runs/34931461950) probes `e2e-manifest-extract-isolation: ok (28s)`, against `FAILED` before.

Full per-suite log, which is the part worth reading — a suite can go green by skipping everything, and this one does not:

```
SKIP: signal-death check (an APE's exit status is unreadable from this shell)
SKIP: failure-legibility check (this hull cannot link an app here)
PASS: child protocol: 'ok' carries the manifest JSON back
PASS: child protocol: 'none' distinguishes a manifest-less app from a failure
PASS: child protocol: 'err' reports a failure with a message
PASS: child protocol: the entry marker is replaced by the real result
SKIP: a JS app with a manifest (this hull cannot link an app here)
SKIP: a manifest-less JS app (this hull cannot link an app here)
SKIP: isolated-child spawn check (this hull cannot link an app here)
e2e_manifest_extract_isolation: ALL PASS
```

Checks 2a–2d are the substance — they are what proves isolation is really engaged rather than inferred from an outcome — and all four assert for real here, on recovered statuses.

Reading that log did expose one thing I had got wrong, fixed in fecc16fe: it invites the conclusion that check 1 did nothing. It did — the status-free half ran on all 10 runs and is what would still catch a build wrongly succeeding into an artefact. It just printed nothing, because the skip replaced the only line it had. Both are now reported.

PR #506, opened 2026-09-15
test(windows): e2e_compute_headers — two refusal checks read a status the shell cannot see

farkasmark:
Next suite in the Windows e2e migration. Probed verdict `FAILED`, 7 passed / 2 failed:

```
FAIL: refresh: should fail when a header cannot be staged (rc=0)
FAIL: new: should refuse an existing module dir
```

Both refusals **happened**. Both were reported as defects because the shell read `rc=0` — on Windows an APE reports its exit status shifted left by 8 and a POSIX shell keeps the low byte (jart/cosmopolitan#1521).

These are the only two checks in the suite that turn on a status; the other seven assert on file contents and were already correct everywhere. So the fix is exactly those two, through `hull_rc`.

### One thing verified rather than assumed

`hull_rc` drives the binary through `Start-Process`, and this script `cd`s into its scratch directory and then uses **relative** module paths (`compute new foo`). If `Start-Process` did not inherit the caller's working directory, the conversion would quietly change what the command operates on.

Measured with the same quoting the helper uses:

```
$ cd <scratch>; echo HELLO_MARKER > marker.txt
$ Start-Process findstr.exe -ArgumentList '"HELLO_MARKER"','"marker.txt"' ...
exit 0
stdout: HELLO_MARKER
```

The relative filename resolved in the shell's cwd and a real status came back. Safe here, and worth knowing for the next suite converted.

### Elsewhere

`hull_rc_init` no-ops off Windows, so Linux and macOS take the direct call and assert exactly what they did before.


farkasmark:
Green on Windows: [run 34931665236](https://github.com/artalis-io/hull/actions/runs/34931665236) probes `e2e-compute-headers: ok (11s)`, against `7 passed, 2 failed` before.

Worth noting this one goes green with **zero skips** — every check asserts for real on Windows, including the two that were failing:

```
PASS: new: installs both hull_compute.h and hull_span.h
PASS: new: installed hull_span.h is byte-exact to templates/hull_span.h
PASS: refresh: restores BOTH headers to canonical bytes
PASS: refresh: legacy hull_compute.h-only project gains hull_span.h (compute.h intact)
PASS: refresh: idempotent (identical bytes on repeat)
PASS: refresh: fails non-zero when a header cannot be staged
PASS: refresh: rollback - neither header changed on failure (no mismatched pair)
PASS: refresh: no stray .hull-tmp/.hull-bak files after failure
PASS: new: refuses an existing module dir

compute-headers: 9 passed, 0 failed
```

Nothing here needs a linkable hull, so unlike the neighbouring suites there is no unsupported-configuration guard to add — recovering the status was the entire gap. The `Start-Process` cwd measurement in the description is what made that safe to assume for the relative `compute new foo` paths.

PR #507, opened 2026-09-15
test(windows): e2e_context's one exit-code check could not see an exit code

farkasmark:
Next suite in the Windows e2e migration. Probed `FAILED` on a single check:

```
FAIL: missing task exit - expected exit 1, got 0
```

`hull agent context` with no `--task` **did** exit 1. On Windows an APE reports its status shifted left by 8 and a POSIX shell keeps the low byte (jart/cosmopolitan#1521), so the usage error read as a success. Same defect as #503; this is the only status assertion in the file, so it is the whole fix.

The other two `|| EXIT_CODE=$?` sites are deliberately untouched — they never assert on the code, they exist to stop `set -e` aborting on hull's non-zero exit, and they still do that.

### One thing measured rather than assumed

`HULL` here is the relative `./build/hull`, and `hull_rc` drives the binary through `Start-Process`, so a relative program path only works if PowerShell resolves it against the shell's cwd. `cygpath -w ./build/hull` keeps it relative (`build\hull`), so this mattered:

```
$ Start-Process -FilePath 'sub\f.exe' -ArgumentList '"HELLO_MARKER"','"marker.txt"' ...
exit 0
stdout: HELLO_MARKER
```

It resolves, and a real status comes back — so there was no reason to disturb the `HULL` assignment.


farkasmark:
Green on Windows: [run 34936600418](https://github.com/artalis-io/hull/actions/runs/34936600418) probes `e2e-context: ok (8s)`, against `FAILED` before.

```
PASS: missing task exit
57/57 e2e context tests passed
```

**57/57, no skips** — the check asserts for real on Windows rather than being stood down, which is the outcome worth having where the platform allows it.

PR #508, opened 2026-09-15
test(windows): e2e_attachment — a locale-sensitive id extraction, found by making the suite talk

farkasmark:
**Superseded its own first commit.** 6490c37a is reverted in 07d7c8a4; read this description, not that one.

## What I got wrong

6490c37a claimed a native curl cannot *open* a UTF-8-named file but can *send* a UTF-8 filename, so declaring the name via `;filename=` would fix the check. Measured against a dead port, that held:

```
rc=7   -F file=@img.png;filename=<utf8 name>
rc=26  -F file=@<utf8 name>
```

It still failed identically on the runner — `Error 3`, same place, [run 34937553613](https://github.com/artalis-io/hull/actions/runs/34937553613) against the branch head. Two things were wrong with it: the measurement was taken in **Git Bash (MINGW mode)** while the runner is **MSYS2 in MSYS mode** — a different argv-conversion and locale layer — and `rc=3` is `CURLE_URL_MALFORMAT`, which never matched a file-open failure in the first place. Reverted rather than left in the tree looking fixed.

## What this PR now does

The part that *is* supportable on the evidence to hand: the reason nobody could tell what failed.

`set -e` plus a bare command substitution killed the script at the assignment, so the suite died with a bare `make: *** Error 3` after 22 passing assertions and **no FAIL line** — the one shape a reader cannot classify from a probe tail. Its own graceful failure path, which already knew how to report this, was never reached.

curl's status and message are captured, so that path runs:

- the check fails with `curl rc=N resp: ...` instead of killing the suite,
- the JS half runs instead of being lost,
- the next probe says what the actual failure is rather than costing another round trip to ask.

The suite will still go red on Windows — correctly, because this check genuinely does not work there yet. It will now say why.


farkasmark:
Green on Windows: [run 34941742102](https://github.com/artalis-io/hull/actions/runs/34941742102) probes `e2e-attachment: ok (6s)` — **74 passed, 0 failed**, both runtimes, no skips.

The ten RFC 5987 assertions have never run on Windows before, and all pass:

```
PASS: lua unicode: é encoded      PASS: js unicode: é encoded
PASS: lua unicode: 文 encoded     PASS: js unicode: 文 encoded
PASS: lua unicode: 档 encoded     PASS: js unicode: 档 encoded
PASS: lua unicode: 📄 encoded     PASS: js unicode: 📄 encoded
PASS: lua unicode: ASCII fallback PASS: js unicode: ASCII fallback
```

So Hull's `Content-Disposition` encoding was correct on Windows the whole time. The suite could not see it.

## How this went, since the history is worth reading before the diff

| commit | what it claimed | outcome |
|---|---|---|
| 6490c37a | a native curl cannot open a UTF-8-named file | **wrong** — reverted |
| 07d7c8a4 | guard the upload so the failure is legible | right idea, wrong command |
| 2bd9aac9 | guard the serve curl too | that was the one dying |
| 1bb2158b | `LC_ALL=C` on the id extraction | green |

Two wrong calls, and neither was broken by a better guess. What broke it was making the suite **report** instead of inferring which line had failed. Once both curls were guarded it printed:

```
curl: (3) unmatched close brace/bracket in URL position 84:
.../serve/allow/7546490ce34ead41cde6926027de4f2a<bad>.png","size":29}]}
```

which names the real defect at once: the `sed` extraction's trailing `.*` has to consume the rest of the JSON — including the filename's bytes — and where it stops at a byte it cannot decode, the tail survives into the captured id. A 32-hex id with JSON glued to it then builds a malformed URL.

All three id extractions get `LC_ALL=C`, not just the one that failed, so they cannot drift apart. Byte-wise matching is what pulling a hex id out of a byte stream wants on any platform.

## What stays

The guards. A bare `make: *** Error 3` after 22 passing assertions and no FAIL line is the one shape a reader cannot classify, and this suite produced exactly that for as long as it has run on this platform. It now reports whatever happens, on every host.

The reverted theory is recorded rather than buried — 07d7c8a4's message says what 6490c37a claimed, which measurement supported it, and why that measurement did not transfer: it was taken in Git Bash (MINGW mode), and the runner is MSYS2 in MSYS mode.

farkasmark:
Also confirmed green in a later batch alongside the others.

PR #510, opened 2026-09-15
test(windows): e2e_build_flavor's cosmo skip never fired, so it ran where it must not

farkasmark:
```
FAIL: pure-compute app should exit 7, got 127
```

Nothing exits 127 here. That is `hull_rc` reporting there was **no binary to run**: the `hull build` two lines above had failed, unnoticed, because that check reads an APE's exit status and cannot see one on Windows.

But the build failing is not the defect either.

## The actual defect

This suite is **cosmo-exempt** — a fat APE keeps every composable subsystem in-base, so there is no flavor axis to test — and it already carried a skip saying so. The skip matched `file(1)` output against `*cosmo*|*APE*`, and `file(1)` on Windows calls an APE a `DOS/MBR boot sector`. No branch matched. So the suite ran, on a hull that cannot link the app it then tries to execute.

#501 found this and fixed the identical probe in the five `feature-*` suites by inlining a magic-byte check. **This suite carried the same probe and was missed.** `hull_is_ape` is that fix factored into `tests/lib/hull_rc.sh`, so the next suite needing it doesn't have to rediscover why `file(1)` won't do — and `hull_rc_init` now shares the one probe instead of keeping a second copy of it.

## Also

Two `trap ... EXIT` lines, where the second **replaces** the first rather than adding to it, so `$HULL_RC_TMP` was never cleaned up. Same mistake a1cfce6b fixed in `e2e_modular_resolution`, made again here. Folded into one.

## Why the skip is the right answer and not a retreat

The five `feature-*` suites reach the same verdict for the same architectural reason, and their probe output says so explicitly:

```
SKIP: cosmo keeps WAMR in-base (a fat APE can't force-load a feature archive)
SKIP: cosmo keeps SQLite in-base (a fat APE can't force-load a feature archive)
...
```

Those five are green on Windows but contribute **no coverage**, and this one joins them. That is worth stating plainly rather than counting as a win: the composable-feature axis does not exist on a cosmo build, so there is nothing here for a Windows runner to verify.


farkasmark:
Green on Windows: [run 34946029786](https://github.com/artalis-io/hull/actions/runs/34946029786) probes `e2e-build-flavor: ok (2s)`, down from `FAILED (9s)`. The whole log:

```
SKIP: cosmo keeps everything in-base (fat APE)
```

Which is the correct outcome, and the one the suite was already trying to reach — the skip simply never fired. It now contributes no Windows coverage, same as the five `feature-*` suites, and for the same architectural reason.

PR #531, opened 2026-09-17
fix(db): WAL is a performance choice, not a startup requirement

farkasmark:
Resolves the concurrent-startup failure that #530 only narrowed. Of 8 hull processes started together against one database, **1 served; now 8 of 8 do** — verified three consecutive times.

### The processes were never failing to open the database

They opened it fine, then **aborted** — because `hl_cap_db_init` returned -1 if *any* pragma failed, and `PRAGMA journal_mode=WAL` takes an exclusive lock that several processes converting the same fresh database contend for. The loser gets `SQLITE_PROTOCOL`, which is **not a busy condition**: `busy_timeout` doesn't cover it and retrying doesn't clear it (tried during #530's investigation and rejected — it changed nothing and made the failure silent).

So a **tuning pragma losing a race killed the application.**

### Pragmas classified by what a failure means

| | | on failure |
|---|---|---|
| **Required** | `busy_timeout` (first, so later pragmas wait rather than fail instantly), `foreign_keys` (semantic) | **abort** |
| **WAL** | `journal_mode=WAL` | warn, continue |
| **WAL-only** | `synchronous=NORMAL`, `wal_autocheckpoint` | skipped unless WAL took |
| **Tuning** | `cache_size`, `temp_store`, `mmap_size` | ignored |

`foreign_keys` stays fatal because an app told its referential integrity is enforced must not run if it isn't.

`synchronous=NORMAL` is applied **only** if WAL took: it is documented as safe *because* WAL protects against corruption, so without WAL that reasoning collapses. A connection that lost the race keeps the default `FULL` rather than quietly trading durability it was never asked to trade.

### Measured

Windows, one database, distinct ports and directory per case so neither `TIME_WAIT` nor a shared file could confound it:

```
N=1  1/1      N=2  2/2      N=4  4/4      N=8  8/8, 8/8, 8/8
```

The losers still lose the race — and now say so instead of dying:

```
hull: sqlite could not set WAL (locking protocol); continuing with the
database's current journal mode
```

(Deliberately worded that way: journal mode is a property of the **file**, so once the winner converts it, a connection whose own pragma failed is still using WAL. It failed to *set* the mode, not to get it.)

### Not Windows-specific in principle

The failure mode was always there. Windows surfaced it because its SQLite VFS is stricter about this locking window, and because hull's default DSN is the **relative** path `data.db` (`serve.c`) — so processes sharing a working directory share a database without anyone choosing to.

### Follow-up this implies

`e2e-cache-concurrent` was moved to KNOWN in #529 because it failed on Windows. With this fix it should pass, and the tier's `IMPROVED` check will fail the job to force its promotion back to enforced — the mechanism working as designed, but it needs a follow-up PR rather than silently going green.


farkasmark:
**Verified on the Windows runner with the real suite** (not just the diagnostic), and the promotion is folded in:

```
PASS: all 8 workers actually started (8/8)
PASS: all 8 mixed-runtime workers actually started (8/8)
PASS: JS cache also populated under mixed load (2)
11/11 e2e concurrent-cache tests passed
```

`JS cache not populated under mixed load` was the assertion that started this whole thread — it passes for the right reason now: the workers that used to die at startup are running.

`e2e-cache-concurrent` therefore returns to **ENFORCED** in this same PR (89 enforced, 2 known). Splitting it out would have guaranteed one red nightly for nothing: the tier's `IMPROVED` check fails the job the moment a KNOWN suite starts passing, which is precisely what this fix causes. The check did its job — there's just no value in letting it fire when the cause is fixed in the same commit.

PR #540, opened 2026-09-18
ci(windows): install the cosmocc bundle, so hull build can run at all

farkasmark:
`e2e-build` skipped its entire second half on Windows — Steps 4–18, the build/sign/verify/inspect/eject pipeline — because hull reported no compiler it could drive. The `cosmocc.zip` driver already on PATH is a `#!/bin/sh` script, and a cosmo APE cannot exec a shebang; it fails even given an absolute path that exists.

The hull-managed bundle exists for exactly this: it ships `bin/busybox.exe` beside the driver so a cosmo hull can spawn it.

### Result

```
before:  22 pass,  0 fail   (Steps 4-18 skipped entirely)
after:   60 pass, 42 fail   (Steps 4-18 RUNNING for the first time)
```

The chain works end to end: `--tag=` installs the bundle (#536) → hull's own extractor preserves busybox and 38 symlinks where MSYS `tar` fails → doctor reports `build_compiler: cosmocc` (#539's ordering fix) → the suite sets `CAN_BUILD=1`.

The remaining 42 failures share one root cause: hull compiles, reaches the link, produces nothing. Being diagnosed; likely the fat-link shape, since Step 0 runs `make platform` + `EMBED_PLATFORM=1` (native, single-arch) while a cosmo link needs `platform-cosmo` and `EMBED_PLATFORM=cosmo`.

### Not cached, deliberately

401 MB in ~21s from GitHub's CDN. At that speed a cache buys seconds and adds a failure mode — the zig caching attempt on this repo broke `e2e-musl-cross` twice for a download that takes less time than this one.

### Also here

- **Install ordered before the PROBE step**, not just the tier. My first attempt put it above the enforced tier only, and the probe step sits ~200 lines earlier — the timestamps showed `e2e-build` running 17:13–17:22 and the install landing at 17:22:28.
- **`check_exit` now prints the command's output on failure.** It reported `expected exit 0, got 1` while the linker error sat in `$OUT`, one variable away. Same defect that cost five probes on the SQLite investigation.
- **The NOT-GATED comment rewritten from the derived set.** It had drifted three ways: described `e2e-build` as dying in make's job control (disproved in #527), cited the 50-minute cap (raised to 65 in #526), and listed `e2e-cache-cosmo` as ungated (gated in #526).

`e2e-build` is **not** gated by this PR. That waits on a run showing Steps 4–18 actually pass.


farkasmark:
Superseded by #541, which keeps the two changes that stand on their own and drops the rest.

**The premise behind this PR was wrong.** `hull build` on Windows is not untested — the separate `windows-app-build` job fetches cosmocc *with busybox*, builds the multi-arch platform, builds hull with it embedded, then runs `hull new` + `hull build` and **executes the produced APE**. `e2e-build`'s skip is a duplication gap, not a coverage hole.

**And the two halves here are coupled, each bad alone.** Installing the bundle by itself flips `CAN_BUILD=1` and turns an honest skip into 42 failures; adding the `make platform-cosmo` Step 0 that would fix those ran 1200s without finishing — which is precisely the cost `windows-app-build` exists to isolate (*"three builds where the sibling job does one"*).

#541 keeps `check_exit` printing the output on failure (which is what made the 42 failures collapse to one readable line) and the corrected NOT-GATED comment, and records what was tried here so it isn't repeated.

PR #543, opened 2026-09-19
chore(keel): upgrade the vendored Keel to v3.1.1

farkasmark:
> **Do not merge as-is.** The upgrade is sound on API grounds and green everywhere except Windows, where it costs **53ms on every hull process start**. Opened as the record of that finding.

Three releases on from the `v3.0.0` pin (3.0.1, 3.1.0, 3.1.1) — 51 commits.

## API impact: nil in practice

Five public headers changed, all additive except one signature:

```c
kl_http_response_file(res, KlSocketHandle fd, size)  ->  (res, int fd, size)
```

Hull never calls it — it appears only in two historical comments in `static.c`. Keel documents it as source-compatible, and it *restores* the original declaration that the 3.0 socket-handle retype swept up by mistake.

**An API reading would have shipped this.** That is the point of what follows.

## The regression

`hull version`, 40 consecutive starts, Keel pin the only variable:

| | v3.0.0 | v3.1.1 | |
|---|---|---|---|
| `hull version` | **21ms** | **74ms** | **3.5×** |
| `hull --help` | 21ms | 31ms | 1.5× |

+53ms per process start. It is not confined to CI — every `hull` invocation on Windows pays it.

### What it does to the enforced tier

| suite | pre | post | |
|---|---|---|---|
| `e2e-analyze` | ok 33s | **FAIL 300s** (timeout) | +267s |
| `e2e-cache-module` | ok 93s | ok 217s | +124s |
| `e2e-cache-cosmo` | ok 810s | **FAIL 900s** (timeout) | +90s |
| ~85 others | — | — | +2…10s |

Only suites that spawn many hull processes moved. `e2e-analyze`'s +267s is ≈5000 invocations × 53ms, which is the whole of it.

(`e2e-cache-concurrent` 31s→51s is unrelated — that suite now waits for 8 workers to actually boot instead of killing them at 2s, from #529.)

## Likely cause — measured vs inferred

**Measured:** startup cost 21ms → 74ms, Keel pin the only change.

**Inferred, from Keel's own changelog:** 3.1.0 replaced the load-time constructor that starts Winsock with a constructor-free `InitOnceExecuteOnce` path in the PAL, *because MSVC has no equivalent of the constructor*. That is a per-process startup path on Windows, and it was written for MSVC and MinGW — **a Cosmopolitan APE is neither.**

I have not isolated it further (e.g. against an `HL_ENABLE_HTTP=0` build that links no Keel at all), so treat the mechanism as strongly indicated rather than proven.

## Other CI surfaces

- Linux / macOS matrix: **fully green**
- Windows `hull built on Windows builds + runs an app`: **green**
- Windows enforced tier: **fails**, as above

## Suggested disposition

Report upstream to Keel with the measurement and hold the bump. Raising the two timeouts would hide a 3.5× startup regression that users feel on every command, which is the wrong trade. It is Keel's change and Keel's fix; Hull is a legitimate consumer (cosmo APE) that neither MSVC nor MinGW testing covers.


farkasmark:
**Taking the bump, with the regression reported rather than hidden.**

Upstream: artalis-io/keel#328 — +53ms on every hull process start on Windows (21ms → 74ms, Keel pin the only variable), from 3.1.0 replacing Winsock's load-time constructor with a per-process `InitOnceExecuteOnce` path written for MSVC and MinGW.

Paired with timeout accommodations so the nightly isn't permanently red:

| | was | now |
|---|---|---|
| `e2e-analyze` | 300s | 600s |
| `e2e-cache-module` | 300s | 600s |
| `e2e-cache-cosmo` | 900s | 1200s |
| job cap | 65 min | 80 min |

Each carries its pre-keel#328 value and an instruction to revert when that lands. They accommodate a **regression**, not a genuine cost — and if the timings don't come back down after the upstream fix, the cause was something other than the Winsock change, which is worth discovering.

What I deliberately didn't do: fold the new numbers into the defaults. A budget that quietly absorbs a 3.5× startup regression outlives anyone's memory of why it was set.

